PR #2, opened 2020-12-14
[WIP] Port Bloop's DAP implementation

er1c:
- [x] Copy Bloop DAP code and get it compiling
- [ ] Refactor / deal with copied util files
- [ ] Review library API
- [ ] Update get tests working


build.sbt
```diff
@@ -0,0 +1,25 @@
...
+    libraryDependencies ++= List(
+      "org.scala-lang" % "scala-compiler" % scalaVersion.value,
+      Dependencies.asm,
+      Dependencies.asmUtil,
+      Dependencies.bsp4j,
+      Dependencies.bsp4s,
+      Dependencies.javaDebug,
+      Dependencies.monix,
+      Dependencies.sbtTestInterface,
+      Dependencies.scalameta,
```
tgodzik, line 22: Do we need scalameta actually?

core/src/main/scala/bloop/reporter/Problem.scala
```diff
@@ -0,0 +1,40 @@
+package bloop.reporter
+
+import xsbti.Severity
+
+/** Describes a problem (error, warning, message, etc.) given to the reporter. */
+final case class Problem private (
```
tgodzik, line 6: Anything related to compilation errors should not be needed (I hope). We should probably have separate interfaces here that Bloop can reuse?

tgodzik (commented):
Left some comments for the WIP. Great to see the progress you are making! Let me know if you need help figuring things out.

build.sbt
er1c: I added scalameta for the `scala.meta.io`, there was `AbsolutePath` and something else, but I'll need to double check those usages

er1c:
We need to start reviewing some of the public library API.

```
abstract class DebuggeeRunner {
  def logger: Logger
  def run(logger: DebugSessionLogger): Task[ExitStatus]
  def classFilesMappedTo(origin: Path, lines: Array[Int], columns: Array[Int]): List[Path]
}
```

I was thinking  of adding:

```
trait ExitStatus {
  def code: Int
  def name:  Int
}
```

It looks like bloop and sbt would support `xsbt.Logger`

build.sbt
ckipp01: I hit on this same exact thing not that long ago in a project. If all you need in `AbsolutePath` or some other path related util, it might just be best to replace that. If not, you're bringing in a whole lot of stuff with scalameta that won't be needed.

er1c: might want `import scala.meta.jsonrpc.{JsonRpcClient, Response => JsonRpcResponse,` so  we'll see.

core/src/main/scala/dap/BspServerContext.scala
```diff
@@ -0,0 +1,201 @@
...
+                      val listenAndUnsubscribe = startedServer.listen
+                        .runOnComplete(_ => backgroundDebugServers -= startedServer)(ioScheduler)
+                      backgroundDebugServers += startedServer -> listenAndUnsubscribe
+
+                      startedServer.address.map {
+                        case Some(uri) => Right(new bsp.DebugSessionAddress(uri.toString))
+                        case None =>
+                          val error = JsonRpcResponse.internalError("Failed to start debug server")
+                          Left(error)
+                      }.runAsync(ioScheduler)
```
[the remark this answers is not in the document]
adpi2: It should take as parameter the debuggeeRunner (that encapsulates the abstract notion of Project).

Which means that only the DebuggeeRunner interface should live in this repository. The implementations will  be provided by Bloop and sbt.

core/src/main/scala/dap/DebuggeeRunner.scala
```diff
@@ -0,0 +1,49 @@
...
+import bloop.cli.ExitStatus
+import monix.eval.Task
+import java.nio.file.Path
+import xsbti.Logger
+import ch.epfl.scala.bsp.ScalaMainClass
+import xsbti.compile.analysis.SourceInfo
+import sbt.internal.inc.Analysis
+import scala.concurrent.Future
+
+sealed abstract class DebuggeeRunner {
```
adpi2, line 12: We need to keep this interface

core/src/main/scala/dap/DebuggeeRunner.scala
```diff
@@ -0,0 +1,49 @@
...
+
+}
+
+abstract class TestSuiteDebugAdapter(filters: List[String]) extends DebuggeeRunner {
+
+}
+
+abstract class AttachRemoteDebugAdapter extends DebuggeeRunner {
+  override def run(logger: DebugSessionLogger): Future[ExitStatus] = Future.successful(ExitStatus.Ok)
+}
```
adpi2, line 49: But maybe these sub-classes aren't useful here.

core/src/main/scala/dap/DebuggeeRunnerFactory.scala
```diff
@@ -0,0 +1,46 @@
+package dap
+
+import ch.epfl.scala.bsp
+
+trait DebuggeeRunnerFactory[Project, State] {
```
adpi2, line 5: Also I am pretty sure that factory should stay in Bloop

adpi2 (commented):
Thanks @er1c for the effort your are putting in this!

So far it looks very promising.

I added a few comments that I hope will be helpful to you. Don't take them for granted because I may have missed a few important points...

er1c:
@adpi2 I'd say its going, slowly, but making some progress - mostly just understanding.

If you want to add this as a branch to this repo and see if you or @tgodzik are able to clean up parts of it too, I wouldn't mind :)

I'm still trying to figure out all of the interfaces and types, so the code is mostly still play ground. I think the goal is to get enough of it refactored that the bloop part works, and then the sbt part can be figured out separately.

core/src/main/scala/dap/DebuggeeRunnerFactory.scala
er1c: just for reference the `bloop-dap` project is the playground I was using for stuff "in bloop" I was trying to figure out if there is a shared interface, and then the bloop specific factory would be there

build.sbt
```diff
@@ -0,0 +1,61 @@
...
+
+lazy val bloopDap = project
+  .in(file("bloop-dap"))
+  .settings(
+    name := "bloop-scala-debug-adapter",
+    libraryDependencies ++= List(
+      "ch.epfl.scala" %% "bloop-frontend" % bloopVersion
+    )
+  )
+  .dependsOn(core)
```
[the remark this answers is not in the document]
er1c: yes

core/src/main/scala/dap/BspServerContext.scala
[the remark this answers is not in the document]
er1c: I'm still thinking about these comments, I think it makes sense.  I may have been going down a path of over-generalizing

build.sbt
```diff
@@ -0,0 +1,61 @@
...
+      Dependencies.asm,
+      Dependencies.asmUtil,
+      Dependencies.bsp4j,
+      Dependencies.bsp4s,
+      Dependencies.javaDebug,
+      Dependencies.monix,
+      Dependencies.sbtTestInterface,
+      //Dependencies.scalameta,
+      Dependencies.zinc,
+      "ch.epfl.scala" %% "bloop-frontend" % bloopVersion
```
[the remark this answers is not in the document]
er1c: FYI: https://github.com/sbt/sbt/pull/6226 I got the sbt-native debug messages in sbt for use in the plugin.

core/src/main/scala/dap/DebuggeeRunner.scala
er1c: Probably not needed, I  was trying to figure out if there was any shared logic here, so far the blank classes seem to communicate the need :)

core/src/main/scala/dap/DebugSession.scala
```diff
@@ -0,0 +1,371 @@
...
+ *  The JDI address of the debuggee is obtained through the [[DebugSessionLogger]]
+ */
+final class DebugSession(
+    socket: Socket,
+    initialDebugState: DebugSession.State,
+    context: IProviderContext,
+    initialLogger: Logger,
+    ioScheduler: Scheduler,
+    loggerAdapter: DebugSession.LoggerAdapter
+) extends DapServer(
```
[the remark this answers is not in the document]
er1c: What is a reasonable type signature for 

```
abstract class DebuggeeRunner {
  def logger: Logger
  def run(logger: DebugSessionLogger): Future[DebuggeeExitStatus]
  def classFilesMappedTo(origin: Path, lines: Array[Int], columns: Array[Int]): List[Path]
}
```

`Future` is generic, but also does't support the cancel / on finish like:

```
    val startAndListen = listen(handle.server)
      .doOnFinish(closeServer)
      .doOnCancel(closeServer(None))
```

should `DebugServer` be an interface? Similiar  issue with `DebugSession`:

```
    debugState.transform {
      case DebugSession.Ready(_) =>
        socket.close()
        DebugSession.Cancelled

      case DebugSession.Started(debuggee) =>
        cancelPromises(new CancellationException("Debug session cancelled"))
        cancelDebuggee(debuggee)
        scheduleForcedEndOfConnection()
        DebugSession.Cancelled

      case DebugSession.Cancelled => DebugSession.Cancelled
    }
```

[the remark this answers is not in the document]
adpi2: `Future` does not seem to be "powerful" enough because, as you mentionned, it is missing the `onFinish`, the `onCancel` callbacks and the `Cancellable`.

I suggest we take just the things we need from Monix tasks, wich are:
  - `Cancellable` the ability to cancel the `Process`
  - the ability to pass an `onCancel` callback
  - the ability to pass an `onFinish` callback
  
 So `DebuggeeRunner` could be:
 ```scala
 trait class DebuggeeRunner {
  def logger: Logger
  def run(logger: Logger, onCancel:  () => Unit, onFinish: Option[Throwable] => Unit): Cancellable
  def classFilesMappedTo(origin: Path, lines: Array[Int], columns: Array[Int]): List[Path]
}

trait Cancellable {
  def cancel(): Unit
}
```

That interface is easy to implement in Bloop on top of the actual `Task` abstraction. In sbt it should be possible to implement as well with some effort.

 > should DebugServer be an interface? Similiar issue with DebugSession:

No I don't think `DebugSession` or `DebugServer` should be interfaces.

Also I am under the impression that `DebugServer` could expose `Future` instead of `Task`.

tgodzik:
@er1c I can spend some time next week to help out, but don't want to step on your toes. What parts we could help out with or maybe better how could we split the work?

er1c:
https://github.com/scalacenter/scala-debug-adapter/pull/3

PR #4, opened 2021-01-14
[WIP] - Start flushing out sbt-dap-plugin

er1c:


sbt-dap-plugin/src/main/scala/sbtdap/SbtDebugAdapterPlugin.scala
```diff
@@ -3,26 +3,64 @@ package sbtdap
...
 
+  private def handleDebugSessionStart(
+    r: JsonRpcRequestMessage,
+    projects: Seq[Project],
+    state: State
+  ): BspResponse[DebugSessionAddress] = {
+    for {
+      params: DebugSessionParams <- getDebugSessionParams(r)
+      runner <- SbtDebuggeeRunner.inferDebuggeeRunner(params, projects, state)
+      // TODO: This needs to be refactored into a Future/callbacks/etc
```
er1c, line 44: Need to get the `DebugServer` and then the URI from that

PR #8, opened 2021-01-22
Integration with Bloop

adpi2:
This PR makes the `scala-debug-adapater` ready for integration with Bloop.

I've created a corresponding PR in Bloop to run the tests: https://github.com/scalacenter/bloop/pull/1444. We should keep it open and run the tests regularly until the core library is ready to be releases.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -10,4 +10,5 @@ object Dependencies {
   val utest = "com.lihaoyi" %% "utest" % "0.6.6"
   val scalaCompiler = "org.scala-lang" % "scala-compiler" % scala212
   val io = "org.scala-sbt" %% "io" % "1.4.0"
+  val bloopFrontend = "ch.epfl.scala" %% "bloop-frontend" % "1.4.5"
 }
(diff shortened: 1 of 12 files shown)


project/Dependencies.scala
```diff
@@ -10,4 +10,5 @@ object Dependencies {
   val utest = "com.lihaoyi" %% "utest" % "0.6.6"
   val scalaCompiler = "org.scala-lang" % "scala-compiler" % scala212
   val io = "org.scala-sbt" %% "io" % "1.4.0"
+  val bloopFrontend = "ch.epfl.scala" %% "bloop-frontend" % "1.4.5"
```
tgodzik, line 13: Should this depend on Bloop?

adpi2: The core module doesn't depend on Bloop.

But we still have a `bloop-dap` module that contains an implementation of `DebuggeeRunner` for Bloop. It's only used to check that the API can be used in Bloop. We can easily remove that module now or later.

er1c (approved):
LGTM, I want to get this merged in, I might add a couple questions after.

PR #9, opened 2021-01-23
Add sbt plugin

adpi2:
Continuation of #4


er1c:
🥳 great progress!!!  I'm hoping to have some time this weekend to hack on this, should we just merge this in now, or did you want to keep it a draft?

sbt-dap-plugin/src/main/contraband/dap.contra
```diff
@@ -0,0 +1,20 @@
...
+  ## The target's Uri
+  uri: java.net.URI!
+}
+
+type DebugSessionParams {
+  ## A sequence of build targets affected by the debugging action.
+  targets: [sbt.internal.bsp.BuildTargetIdentifier]
+
+  ## The kind of data to expect in the `data` field.
+  dataKind: String
```
er1c, line 16: Two things:

1) I had this question back in december in #bloop that I never really got a good response to

image

I *think* having it as an Option is easier for reporting back error messages, but want to just identify that the spec says it should be required, and that this is a conscious engineering decision to make it optional in contraband for our parsing purposes.

2) If we want to target sbt 1.5.x, then we can use the nightly version which already has these available: https://github.com/sbt/sbt/blob/develop/protocol/src/main/contraband/bsp.contra#L66-L80.  For 1.4.x (or other) compatibility, this approach here is probably fine.  Again just wanted to make sure its a conscious decision.

sbt-dap-plugin/src/main/scala/dap/DebugAdapterPlugin.scala
```diff
@@ -0,0 +1,179 @@
...
+        params <- Converter.fromJson[ScalaMainClass](json).toEither.left.map { cause =>
+          Error.invalidParams(s"expected data of kind ${DataKind.ScalaMainClass}: ${cause.getMessage}")
+        }
+      } yield {
+        val forkOptions = ForkOptions(
+          javaHome = javaHome,
+          outputStrategy = None,
+          bootJars = Vector.empty[File],
+          workingDirectory = Option(workingDirectory),
+          runJVMOptions = params.arguments,
```
er1c, line 74: We need to inject `s"-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,quiet=n"` at some point in time

sbt-dap-plugin/src/main/contraband/dap.contra
adpi2: > 1. I had this question back in december in #bloop that I never really got a good response to

I have no clear answer. It seems to be a convention from the BSP protocol, in the [test](https://build-server-protocol.github.io/docs/specification.html#test-request) and [run](https://build-server-protocol.github.io/docs/specification.html#run-request) requests for instance. The none case is never implemented in Bloop because Metals does not need it.

But it could be implemented this way in sbt (similarly in Bloop):
- if the target is a `Compile` target then run the main class
- if the target is a `Test` target then run the test

It is not yet needed, but could be nice to have in the future.

> 2. If we want to target sbt 1.5.x, then we can use the nightly version

Right. Chances are our plugin will be released before the final `1.5.0` version so it's cool to have it working on `1.4.x`. Some time after sbt `1.5.0` is released, we can get rid of our own baked codec.

sbt-dap-plugin/src/main/scala/dap/DebugAdapterPlugin.scala
adpi2: Yes absolutely. It should go to `runJVMOptions`

Also we need to inject the `-classpath` option in the `runJVMOptions`. It can be constructed by resolving `fullClasspath.value`. But more interestingly we can copy the `exportedProducts.value` to our own safe directory (somewhere in `target.value`) and concatenate it to the `dependencyClasspath.value`. So that we can compile safely while the debug session is running.

adpi2:
Merged! Feel free to take over it and to push what you've done.

I'll probably have some time tomorrow as well.

sbt-dap-plugin/src/main/scala/dap/DebugAdapterPlugin.scala
er1c: Part of this is over my head, but I'm glad you are involved :)

PR #10, opened 2021-01-23
Add unicode/microsite for project

er1c:
Started adding some additional boilerplate.


build.sbt
```diff
@@ -47,11 +103,78 @@ lazy val sbtPlugin = project
...
+lazy val site = project.in(file("site"))
+  //.disablePlugins(MimaPlugin)
+  .enablePlugins(MicrositesPlugin)
+  .enablePlugins(MdocPlugin)
+  .settings(doNotPublishArtifact)
+  .dependsOn(core)
+  .settings {
+    import microsites._
+    Seq(
+      micrositeName := projectTitle.value,
```
tgodzik, line 123: I'm thinking that maybe we could use the same as we use for Metals with docusaurus? It seems much more simple:
https://github.com/scalameta/metals/blob/main/build.sbt#L690

It's mostly just `.md` files and super easy to deploy.

er1c: Cool, I'll take a look!

er1c:
going to replace with mdoc

PR #11, opened 2021-01-24
sbt plugin WIP

er1c:
Not entirely sure the return value is correct, I got the tip on analysis from: https://twitter.com/eed3si9n/status/1353456684430225408


adpi2 (approved):
This is looking very good! Thanks!

PR #12, opened 2021-01-25
Add classpath and debug option

adpi2:
Run scripted tests in CI

Also I have abandoned the idea of using the sbt `State` to store the debug servers. I now think it is unnecessary complexity.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -43,10 +43,13 @@ lazy val sbtPlugin = project
   .in(file("sbt-dap-plugin"))
   .enablePlugins(SbtPlugin, ContrabandPlugin, JsonCodecPlugin)
   .settings(
-    name := "sbt-dap-plugin",
+    name := "sbt-debug-adapter",
     Compile / generateContrabands / contrabandFormatsForType := ContrabandConfig.getFormats,
     scriptedLaunchOpts += s"-Dplugin.version=${version.value}",
-    scriptedBufferLog := false,
+    scriptedDependencies := {
+      val r1 = publishLocal.value
+      val r2 = (core / publishLocal).value
+    }
   )
   .dependsOn(core)
 
--- a/sbt-dap-plugin/src/main/scala/dap/DebugAdapterPlugin.scala
+++ b/sbt-dap-plugin/src/main/scala/dap/DebugAdapterPlugin.scala
@@ -13,24 +13,24 @@ import sjsonnew.support.scalajson.unsafe.{CompactPrinter, Converter, Parser => J
 import scala.concurrent.ExecutionContext
 import java.io.File
 import dap.{DebugSessionDataKind => DataKind}
-
-
+import dap.codec.JsonProtocol._
+import sbt.internal.bsp.codec.JsonProtocol._
+import scala.collection.mutable
 object DebugAdapterPlugin extends sbt.AutoPlugin {
   import SbtLoggerAdapter._
   private final val DebugSessionStart: String = "debugSession/start"
 
-  object autoImport {
-    val dapMainClassSession = inputKey[StateTransform]("Start a debug session for running a scala main class").withRank(KeyRanks.DTask)
-    val dapTestSuitesSession = inputKey[StateTransform]("Start a debug session for running test suites").withRank(KeyRanks.DTask)
-    val dapAttachRemoteSession = inputKey[StateTransform]("Start a debug session on a remote process").withRank(KeyRanks.DTask)
+  // each build target can only have one debug server
+  private val debugServers = mutable.Map[BuildTargetIdentifier, DebugServer]()
 
-    // attribute key for storing the instance of [[DebugServer]]
-    // each build target can have only one buildServer
-    val debugServers = AttributeKey[Map[BuildTargetIdentifier, DebugServer]]("debugServers")
+  object autoImport {
+    val startMainClassDebugSession = inputKey[URI]("Start a debug session for running a scala main class").withRank(KeyRanks.DTask)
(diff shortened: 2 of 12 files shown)


build.sbt
```diff
@@ -43,10 +43,13 @@ lazy val sbtPlugin = project
   .in(file("sbt-dap-plugin"))
   .enablePlugins(SbtPlugin, ContrabandPlugin, JsonCodecPlugin)
   .settings(
-    name := "sbt-dap-plugin",
+    name := "sbt-debug-adapter",
```
er1c, line 46: I like this, should we just git mv the `sbt-dap-plugin` source tree to `sbt-debug-adapter`?

sbt-dap-plugin/src/main/scala/dap/DebugAdapterPlugin.scala
```diff
@@ -40,6 +40,8 @@ object DebugAdapterPlugin extends sbt.AutoPlugin {
   private implicit val executionContext =
     ExecutionContext.fromExecutor(DebugServerThreadPool.executor)
 
+  override def trigger = allRequirements
```
er1c, line 43: cool, I had to look this up, still a little vague syntax but 👍🏻

sbt-dap-plugin/src/main/scala/dap/DebugAdapterPlugin.scala
```diff
@@ -139,17 +107,67 @@ object DebugAdapterPlugin extends sbt.AutoPlugin {
...
-    StateTransform { state =>
-      val allServers = state.get(debugServers).getOrElse(Map())
-      // if there is a server for this target then close it
-      allServers.get(targetId).foreach(_.close())
-      val server = DebugServer(runner, logger)
-      server.start()
-      // not sure if it's possible to respond in a state event
-      state.respondEvent(DebugSessionAddress(server.uri))
-      state.put(debugServers, allServers.updated(targetId, server))
+  private def startMainClassTask: Def.Initialize[InputTask[URI]] = Def.inputTask {
```
er1c, line 110: awesome sauce, and should be easy for me to add the test debugger based upon this

sbt-dap-plugin/src/main/scala/dap/SbtDebuggeeRunner.scala
```diff
@@ -8,7 +8,7 @@ import xsbti.FileConverter
 import sbt.internal.inc.Analysis
 import sbt.internal.inc.SourceInfos
 
-private abstract class SbtDebuggeeRunner(analyses: Array[Analysis], converter: FileConverter, sbtLog: sbt.Logger) extends DebuggeeRunner {
+private abstract class SbtDebuggeeRunner(analyses: Seq[Analysis], converter: FileConverter, sbtLog: sbt.Logger) extends DebuggeeRunner {
   import SbtLoggerAdapter._
```
er1c, line 12: My implicit class for `SbtLoggerAdapter` felt like a bit of a hack, but I didn't want to put time into something else, do you think there is/was a better approach?

sbt-dap-plugin/src/sbt-test/main/breakpoints/test
```diff
@@ -1 +1,2 @@
-> clean; compile
\ No newline at end of file
+> 'Compile / startMainClassDebugSession {"class":"BreakpointTest","arguments":[],"jvmOptions":[]}'
```
er1c, line 1: woot!

er1c (approved):
Looking good, thanks again Obi Wan Adrien!

build.sbt
adpi2: Yep sure!

sbt-dap-plugin/src/main/scala/dap/DebugAdapterPlugin.scala
adpi2: From what I have understood today this `trigger` field is needed to enable the plugin to all projects automatically.

sbt-dap-plugin/src/main/scala/dap/DebugAdapterPlugin.scala
adpi2: For the tests session we probably need to mix our sauce with a bit of these ingredients:
- https://github.com/sbt/sbt/blob/0a6e5c311696ae2ae39e3da2ac80f9206a64179d/main/src/main/scala/sbt/Defaults.scala#L1478-L1489
- https://github.com/sbt/sbt/blob/c4c88b75e448851720ccf8cfc54191a4d28f48a7/main-actions/src/main/scala/sbt/ForkTests.scala#L33-L46

sbt-dap-plugin/src/main/scala/dap/SbtDebuggeeRunner.scala
adpi2: I am totally fine with it.

But maybe it was not a good idea of mine to introduce our own `dap.Logger` interface. All the potential user of the `scala-debug-adpater` has `sbt.Logger` in its classpath already, so we can backtrack to using `sbt.Logger` instead of `dap.Logger` in the core.

er1c: oh right, the bloop extends one of those, but that won't be nice for Mill/etc

adpi2: AFAIK Mill also uses Zinc for compiling Scala and Zinc is part of the sbt ecosystem, it uses `sbt.Logger`.

Also the runners use the Zinc analyses to map sources to class files. So I think our client already need that dependency toward Zinc and it's okay to ask for an instance of `sbt.Logger`.

PR #21, opened 2021-02-05
Use test-client in scripted debugSession/main

er1c:
testing if this fixes CI, might need to enable the plugin from our plugin if thats possible


adpi2:
That's a nice try!

PR #27, opened 2021-02-10
move to scaladebug package

adpi2:


sbt/test-agent/src/main/java/ch/epfl/scala/debugadapter/sbtplugin/internal/SubclassFingerscan.java
```diff
@@ -1,4 +1,11 @@
-package ch.epfl.scala.debug.internal;
+/*
```
er1c, line 1: good find

er1c (approved):
LGTM

Issue #34, opened 2021-02-18
Code Evaluation in Scala 2


adpi2:
1. Implement the `EvaluationProvider` object in the [`DebugAdapter`](https://github.com/scalacenter/scala-debug-adapter/blob/main/core/src/main/scala/ch/epfl/scala/debugadapter/internal/DebugAdapter.scala) class.

The ability to compile an expression should probably be provided by the [`DebuggeeRunner`](https://github.com/scalacenter/scala-debug-adapter/blob/main/core/src/main/scala/ch/epfl/scala/debugadapter/DebuggeeRunner.scala) interface. Also it should be agnostic to the version of the language => no direct dependency to the Scala 2 compiler in the core module.

2. This new method in the `DebuggeeRunner` can then be implemented in the [`SbtDebuggeeRunner`](https://github.com/scalacenter/scala-debug-adapter/blob/main/sbt/plugin/src/main/scala/ch/epfl/scala/debugadapter/sbtplugin/internal/SbtDebuggeeRunner.scala) classes.

tgodzik:
CC @tdudzik  

I wonder if we are actually able to no depend on the Scala compiler. We started looking into this and we could do something like:
- parse an expression using scalameta
- translate the received AST to actual calls to the JDI inteface

The main problem with this approach is implicits, which would need to be explicitly added the the funtion call and without the compiler it might be hard to figure out which implicit needs to be resolved.

The main advantage of this approach on the other hand would be that we could add support for Scala 3 at the same time.

adpi2:
Also there is this implementation for Scala 3: https://github.com/dotty-staging/dotty/blob/debug-server-new/language-server/src/dotty/tools/languageserver/DottyEvaluationProvider.scala

It is using the [dotty.tools.dotc.interactive.InteractiveDriver](https://github.com/lampepfl/dotty/blob/efc9edbff67dd164018fbdc27dba51122fd8149b/compiler/src/dotty/tools/dotc/interactive/InteractiveDriver.scala)

adpi2:
@smarter presented the Dotty Language Server in general and the degbugging support in particular in this talk: https://github.com/lampepfl/dotty/blob/efc9edbff67dd164018fbdc27dba51122fd8149b/compiler/src/dotty/tools/dotc/interactive/InteractiveDriver.scala

smarter:
> @smarter presented the Dotty Language Server in general and the degbugging support in particular in this talk: 

[...]

PR #37, opened 2021-03-04
[PoC] Expression evaluation

tdudzik:
This PR is a basic proof of concept of the expression evaluation. It's based on the idea described here: https://github.com/scalacenter/scala-debug-adapter/issues/34#issuecomment-781333289. 

Implemented features:
- basic arithmetic operations on primitives, variables, and fields,
- invoking (overloaded) methods,
- accessing values,
- basic error handling.

Implemented functionality is obviously far from being complete. Some features can be added rather easily (e.g. better support for primitives and operations on them). However, AST seems to not be enough to implement features like implicits, imports, or creating new instances of objects. It would probably require reimplementing the logic that is already implemented in the compiler. Even some basic operations like `-1.abs` are currently not supported because JDI doesn't provide an access to `Int` object but only to its primitive value. Also, it seems that we don't have access to a `Predef` object which contains some useful functions. There are probably other issues that I didn't describe or haven't discovered yet.

An example of how it works in VSCode:
Screenshot from 2021-03-04 11-05-06

### DIFF
--- a/core/src/main/scala/ch/epfl/scala/debugadapter/internal/EvaluationProvider.scala
+++ b/core/src/main/scala/ch/epfl/scala/debugadapter/internal/EvaluationProvider.scala
@@ -1,0 +1,212 @@
+package ch.epfl.scala.debugadapter.internal
+
+import com.microsoft.java.debug.core.IEvaluatableBreakpoint
+import com.microsoft.java.debug.core.adapter.IEvaluationProvider
+import com.sun.jdi.{DoubleValue, Field, FloatValue, LocalVariable, LongValue, Method, ObjectReference, PrimitiveValue, StackFrame, ThreadReference, Value, VirtualMachine}
+
+import java.util.concurrent.CompletableFuture
+import scala.jdk.CollectionConverters._
+import scala.meta.parsers.Parsed
+import scala.meta.{Stat, Term, _}
+import scala.util.Try
+
+object EvaluationProvider extends IEvaluationProvider {
+  override def isInEvaluation(thread: ThreadReference): Boolean = false
+
+  // Currently, only single expression is supported
+  override def evaluate(
+      expression: String,
+      thread: ThreadReference,
+      depth: Int
+  ): CompletableFuture[Value] = {
+    val frame = thread.frames().get(depth)
+    val result = expression.parse[Stat] match {
+      case Parsed.Success(tree) => evaluate(tree, frame, thread)
+      case Parsed.Error(_, msg, _) => Left(msg)
+    }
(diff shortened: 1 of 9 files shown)


core/src/main/scala/ch/epfl/scala/debugadapter/internal/EvaluationProvider.scala
```diff
@@ -0,0 +1,212 @@
...
+
+  // Currently, only single expression is supported
+  override def evaluate(
+      expression: String,
+      thread: ThreadReference,
+      depth: Int
+  ): CompletableFuture[Value] = {
+    val frame = thread.frames().get(depth)
+    val result = expression.parse[Stat] match {
+      case Parsed.Success(tree) => evaluate(tree, frame, thread)
```
tgodzik, line 24: I wonder if we should actually do the same exact thing, but with the typedTree from the presentation compiler :thinking: 

We should check if we get enough information there. We could modify the source by inserting the evalauted statement at breakpoint, compile it using the existing classpath, get typedTree and translate it to JDI invocations.

That's probably the next thing to check.

tdudzik: If I understand correctly, it would only help us with deciding what should be invoked (e.g. what should be used as an implicit parameter). But the problem is that sometimes invocation is not possible because classes that we need weren't loaded yet.

Example:
```
    val x = 1
    val y = 2
    val z = add(x, y)
    println(z)
  }

  private def add[T](x: T, y: T)(implicit numeric: Numeric[T]) = numeric.plus(x, y)
```
If we set a breakpoint at the line `val z = add(x, y)` and we want to call `add` method, even knowing that we need to use (I think) `scala.math.Numeric.IntIsIntegral` we are not able to do it because this class is not available in JVM (it can be verified using `VirtualMachine.allClasses()`). However, this class will be available if we set a breakpoint at line `println(z)`. I think there are much more cases like that.

tgodzik: We should be able to call the classloader by hand to load up that class. Though, if we already have an implicit parameter then should the class not be loaded already?

tdudzik: `if we already have an implicit parameter then should the class not be loaded already?` I believe classes are loaded lazily, so it's loaded during the first usage.

`We should be able to call the classloader by hand to load up that class` possibility of loading classes would solve multiple problems, so I will try to do it first.

Issue #47, opened 2021-06-15
[Scala 3] Does not stop on some breakpoints


adpi2:
![image](https://user-images.githubusercontent.com/13123162/122014756-c277a280-cdbf-11eb-994c-5cb9f385614d.png)

Problems are:
  - does not stop on line 5
  - breakpoint on line 9 not configured

At first sight, the java debug information seems correct:
```
javap -l App\$package\$
Warning: Binary file App$package$ contains example.App$package$
Compiled from "App.scala"
public final class example.App$package$ implements java.io.Serializable {
  public static final example.App$package$ MODULE$;

  public static {};
    LineNumberTable:
      line 3: 0

  public void start();
    LineNumberTable:
      line 4: 0
      line 5: 3
    LocalVariableTable:
      Start  Length  Slot  Name   Signature
          3       5     1 alice   Ljava/lang/String;
          0       9     0  this   Lexample/App$package$;

  public void greet(java.lang.String);
    LineNumberTable:
      line 8: 0
      line 9: 25
    LocalVariableTable:
      Start  Length  Slot  Name   Signature
         25       7     2 message   Ljava/lang/String;
          0      33     0  this   Lexample/App$package$;
          0      33     1  name   Ljava/lang/String;
}
```

After adding curly braces, everything seems to work:
![image](https://user-images.githubusercontent.com/13123162/122015659-93adfc00-cdc0-11eb-8164-03c9fbd958db.png)

```
javap -l App\$package\$
Warning: Binary file App$package$ contains example.App$package$
Compiled from "App.scala"
public final class example.App$package$ implements java.io.Serializable {
  public static final example.App$package$ MODULE$;

  public static {};
    LineNumberTable:
      line 3: 0

  public void start();
    LineNumberTable:
      line 3: 0
      line 4: 0
      line 5: 3
    LocalVariableTable:
      Start  Length  Slot  Name   Signature
          3       5     1 alice   Ljava/lang/String;
          0       9     0  this   Lexample/App$package$;

  public void greet(java.lang.String);
    LineNumberTable:
      line 7: 0
      line 8: 0
      line 9: 25
    LocalVariableTable:
      Start  Length  Slot  Name   Signature
         25       7     2 message   Ljava/lang/String;
          0      33     0  this   Lexample/App$package$;
          0      33     1  name   Ljava/lang/String;
}
```

tgodzik:
[...]

PR #58, opened 2021-06-18
Fix SourceLookupProvider + add Scala 3 tests

adpi2:


core/src/main/scala/ch/epfl/scala/debugadapter/internal/SourceLookupProvider.scala
```diff
@@ -19,101 +19,52 @@ private[debugadapter] final class SourceLookUpProvider(runner: DebuggeeRunner, l
...
-      } catch {
-        case NonFatal(t) =>
-          logger.error(s"Failed to parse debug line numbers in class file $classFile!")
-          logger.trace(t)
-      }
-    }
-
-    lines
-      .get(line)
-      .orElse {
```
tgodzik, line 86: So the whole fallback here didn't work and was actually causing issues?

adpi2:
I am getting the whole picture now, so I'll try to explain the first part here. The second part will come later in https://github.com/scalacenter/scala-debug-adapter/issues/47.

There are 2 different modes in the scala-debug-adapter to set breakpoints. The first one is to target a source file and the other one is to target a fully qualified class name.

Example:
1. setting breakpoints in the file src/main/example/Hello.scala
```json
{
  "type": "request",
  "seq": 3,
  "command": "setBreakpoints",
  "arguments": {
    "source": {
      "name": "Hello.scala",
      "path": "/home/piquerez/scalacenter/scala-debug-adapter-47/src/main/scala/example/Hello.scala"
    },
    "breakpoints": [{"line": 4},{"line": 5},{"line": 8},{"line": 9}],
    "sourceModified": false
  }
}
```
2. Setting breakpoints in the class `example.Hello$package`
```json
{
  "type": "request",
  "seq": 3,
  "command": "setBreakpoints",
  "arguments": {
    "source": {
      "name": "Hello.scala",
      "path": "dap-fqcn:example.Hello$package"
    },
    "breakpoints": [{"line": 4},{"line": 8}],
    "sourceModified": false
  }
}
```

Mode 1 comes from the java-debug library but it does not work well in Scala because one breakpoint (one line) in the source file can correspond to several breakpoints in different classes. So, in order to make it work, we would have to modify the java-debug library internally.

Currently Mode 1 is not used (by Metals) and the Scala 3 problem does not come from here.

In this PR I just clean Mode 1 a bit so that it can at least pass a simple test. These changes should not affect anything that is being used in Metals or Bloop.

core/src/main/scala/ch/epfl/scala/debugadapter/internal/SourceLookupProvider.scala
adpi2: Not really, this part of the code cannot be reached by a request coming from Metals. This does not fix the Scala 3 bug in Metals.

PR #60, opened 2021-06-18
Fixes #59 - adds sbt-jdi-tools plugin when using java8

### LINKED ISSUE #59: Auto-add in sbt-jdi-tools when using java8
Warning:

```
[warn] The sbt-debug-adapter cannot work because the JDI tools are not loaded.
[warn] 
[warn] Consider adding the sbt-jdi-tools plugin to your `./project/project/plugins.sbt` file:
[warn] addSbtPlugin("org.scala-debugger" % "sbt-jdi-tools" % "x.y.z")
[warn] 
```

Possible solution: https://github.com/ensime/scala-debugger/blob/f8e69614d71786605b4baa8c62f0968683936d22/project/plugins.sbt#L7-L16

Needs to be adapted/tested in the `AutoPlugin` probably.  And version should be `1.1.1`

er1c:
Still need to add a test/manual verification

### DIFF
--- a/sbt/plugin/src/main/scala/ch/epfl/scala/debugadapter/sbtplugin/DebugAdapterPlugin.scala
+++ b/sbt/plugin/src/main/scala/ch/epfl/scala/debugadapter/sbtplugin/DebugAdapterPlugin.scala
@@ -48,7 +48,17 @@ object DebugAdapterPlugin extends sbt.AutoPlugin {
     val startRemoteDebugSession = taskKey[URI]("Start a debug session on a remote process").withRank(KeyRanks.DTask)
     
     val stopDebugSession = taskKey[Unit]("Stop the current debug session").withRank(KeyRanks.DTask)
+
+    {
+      val v = VersionNumber(sys.props("java.specification.version"))
+
+      // If JDK 8 or lower
+      if (v._1.exists(_ == 1) && v._2.exists(_ < 9)) {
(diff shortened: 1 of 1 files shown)


sbt/plugin/src/main/scala/ch/epfl/scala/debugadapter/sbtplugin/DebugAdapterPlugin.scala
```diff
@@ -48,7 +48,17 @@ object DebugAdapterPlugin extends sbt.AutoPlugin {
...
     
     val stopDebugSession = taskKey[Unit]("Stop the current debug session").withRank(KeyRanks.DTask)
+
+    {
+      val v = VersionNumber(sys.props("java.specification.version"))
+
+      // If JDK 8 or lower
+      if (v._1.exists(_ == 1) && v._2.exists(_ < 9)) {
+        addSbtPlugin("org.scala-debugger" % "sbt-jdi-tools" % "1.1.1")
+      }
```
adpi2, line 58: This will add JDI into the user project classpath which is not what we need. We need JDI in sbt own classpath because the debug server runs in sbt's VM.

So basically,
```scala
addSbtPlugin("org.scala-debugger" % "sbt-jdi-tools" % "1.1.1")
```
should appear in `project/project/plugins.sbt` rather than in `project/plugins.sbt`

er1c:
this approach won't work as pointed out

PR #64, opened 2021-07-23
Introduce new `SourceLookUpProvider`

adpi2:
## Background

Currently, the `SourceLookUpProvider` cannot find the source file of a class or the classes of a source file that do not belong to the main build target. As a consequence the scala-debug-adapter cannot set a break point in an internal dependency or a library dependency by itself. So, to make it work, Metals has a [DebugProxy](https://github.com/scalameta/metals/blob/7dc63cbbb5b286012de1079e4d5ae0f34cdf70cf/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProxy.scala) which does the conversion from source file to class names in the [set breakpoint requests](https://github.com/scalameta/metals/blob/7dc63cbbb5b286012de1079e4d5ae0f34cdf70cf/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProxy.scala#L91-L93) and the other way around in the [stack trace responses](https://github.com/scalameta/metals/blob/7dc63cbbb5b286012de1079e4d5ae0f34cdf70cf/metals/src/main/scala/scala/meta/internal/metals/debug/DebugProxy.scala#L108-L119).

So far this was okay, but a problem emerged when we tried to implement the expression evaluation feature, because the scala-debug-adapter needs to read the content of the source file where the expression is evaluated. It has a stack frame number from which it can get a class name but then it doesn't know where the source file is. 

In this PR I propose a new implementation of the `SourceLookUpProvider` that works on the main build target, its internal dependencies (other build targets it depends on) and its library dependencies. This change will make possible the expression evaluation in any source file.

## Implementation Overview

The new `SourceLookUpProvider` is built from a `Seq[ClassPathEntry]` that is provided by the build server (sbt or bloop).

A `ClassPathEntry` is composed of:
- the path of a java class folder or class jar
- its corresponding source entries: a source jar, a source directory or a bunch of source files

For each `ClassPathEntry`, the `SourceLookUpProvider` creates a mapping from class files to source files by reading the source name of each class file. For a given class file we have its file name and we find the actual file using the following rules:
- if there is no source file with that name then it is missing
- if there is only one file with that name it is the right one
- if there is more than one file with that name
  - we choose the one that has the correct path (by convention the path of the source file should be the class package)
  - if none we read the package declarations inside the source files
  - if none the class is said to be orphan

### ClassPathEntryLookUpStats
The [ClassPathEntryLookUpStats](https://github.com/adpi2/scala-debug-adapter/pull/1/files#diff-7fff0cad17594bc138d25f3e032c12bde6e788f840ae7ba9e77a57ab20cc2919) is a test class that also prints some statistics about loading the look-ups of the biggest scala and java libraries. The result are summarized here:

#### Loading time

|                                     | Personal Computer | Github Runner |
|-------------------------------------|-------------------|---------------|
| scala-library-2.13.6                | 167 ms            | 205 ms        |
| scala-compiler-2.13.6               | 340 ms            | 302 ms        |
| scala3-library_3-3.0.0              | 51 ms             | 26 ms         |
| scala3-compiler_3-3.0.0             | 440 ms            | 276 ms        |
| scalatest-core_3-3.2.9              | 148 ms            | 91 ms         |
| cats-core_3-2.6.1                   | 236 ms            | 170 ms        |
| cats-kernel_3-2.6.1                 | 117 ms            | 77 ms         |
| shapeless_2.13-2.3.7                | 224 ms            | 222 ms        |
| org.scala-sbt:main_2.12-1.5.4       | 114 ms            | 94 ms         |
| lm-coursier-shaded_2.12-2.0.8       | 332 ms            | 242 ms        |
| spark-core_2.12-3.1.2               | 370 ms            | 270 ms        |
| spark-catalyst_2.12-3.1.2           | 344 ms            | 274 ms        |
| spark-sql_2.12-3.1.2                | 245 ms            | 202 ms        |
| spark-mllib_2.12-3.1.2              | 182 ms            | 167 ms        |
| akka-actor_2.13-2.6.4               | 161 ms            | 100 ms        |
| akka-stream_2.13-2.6.14             | 207 ms            | 145 ms        |
| akka-http-core_2.13-10.2.4          | 169 ms            | 132 ms        |
| zio_3-1.0.9                         | 92 ms             | 61 ms         |
| zio-streams_3-1.0.9                 | 22 ms             | 14 ms         |
| guava-30.1.1-jre                    | 141 ms            | 103 ms        |
| play_2.13-2.8.8                     | 138 ms            | 129 ms        |
| scalaz-core_2.13-7.3.3              | 501 ms            | 138 ms        |
| sematicdb-scalac-core_2.13.6-4.4.23 | 14 ms             | 8 ms          |

#### Orphan classes

An orphan class is a class file on which we failed to find a corresponding source file. Basically it means it will not be possible to set a break point in this class.

As you can see below the number of orphan classes is low. I suspect this to be better than the current `SetBreakpointsRequestHandler` in Metals but I don't have the corresponding stats for it.

|                                     | Orphan classes |
|-------------------------------------|----------------|
| scala-library-2.13.6                | 0              |
| scala-compiler-2.13.6               | 3 (0.08%)      |
| scala3-library_3-3.0.0              | 0              |
| scala3-compiler_3-3.0.0             | 0              |
| scalatest-core_3-3.2.9              | 0              |
| cats-core_3-2.6.1                   | 0              |
| cats-kernel_3-2.6.1                 | 0              |
| shapeless_2.13-2.3.7                | 0              |
| org.scala-sbt:main_2.12-1.5.4       | 0              |
| lm-coursier-shaded_2.12-2.0.8       | 0              |
| spark-core_2.12-3.1.2               | 1 (0.02%)      |
| spark-catalyst_2.12-3.1.2           | 0              |
| spark-sql_2.12-3.1.2                | 0              |
| spark-mllib_2.12-3.1.2              | 2 (0.08%)      |
| akka-actor_2.13-2.6.4               | 0              |
| akka-stream_2.13-2.6.14             | 0              |
| akka-http-core_2.13-10.2.4          | 0              |
| zio_3-1.0.9                         | 0              |
| zio-streams_3-1.0.9                 | 0              |
| guava-30.1.1-jre                    | 0              |
| play_2.13-2.8.8                     | 2 (0.14%)      |
| scalaz-core_2.13-7.3.3              | 0              |
| sematicdb-scalac-core_2.13.6-4.4.23 | 0              |

### SourceLookUpProviderStats

The [SourceLookUpProviderStats](https://github.com/adpi2/scala-debug-adapter/pull/1/files#diff-a66bf63142bdb2c8396ceedd27a8c8f0114209991606511f16175f5ad90e1b59) is a test class that prints some stats about loading the `SourceLookUpProvider` of a set of dependencies. I choose to load the dependencies of Scaladex because it contains some of the most used Scala libraries: scalatest, akka-http, akka-stream, play, fastparse, json4s, elastic4s...

A total of 115 Scala or Java artifacts are loaded in 1.1 s on my personal computer and 1.9 s on the github runner.

The whole project contains 4 orphan classes (2 in scalatest and 2 in play) which is only 0.012% of all classes.

## Integration with Metals

With this new version of the scala-debug-adapter, Metals does not need to find the class name of a break point in a source file by itself but if it does it would still work. More importantly, the stack trace responses of the debug server now contain the source files so that Metals do not need to convert the class names to source files.

However for the debug server the source files of a library are entries of a source jar file whereas for Metals it's a file in the `.metals/readonly/dependencies` folder. So, with the new version of the scala-debug-adapter Metals only needs to do this conversion: from source jar entry to local dependency source file.

This is already implemented in https://github.com/scalameta/metals/pull/2959

#### SetBreakpoints request

In a `setBreakpoints` request the source field
```json
"source": {
  "name": "list.scala",
  "path": "/home/piquerez/scalacenter/expression-evaluator-test/.metals/readonly/dependencies/cats-core_2.12-2.3.0-sources.jar/cats/instances/list.scala"
}
```
must be converted to
```json
"source": {
  "name": "list.scala",
  "path": "jar:file:///home/piquerez/.cache/coursier/v1/https/repo1.maven.org/maven2/org/typelevel/cats-core_2.12/2.3.0/cats-core_2.12-2.3.0-sources.jar!/cats/instances/list.scala"
}
```

#### StackTrace response

In a `stackTrace` response, the source field of a stack frame:
```json
"source": {
  "name": "list.scala",
  "path": "jar:file:///home/piquerez/.cache/coursier/v1/https/repo1.maven.org/maven2/org/typelevel/cats-core_2.12/2.3.0/cats-core_2.12-2.3.0-sources.jar!/cats/instances/list.scala",
  "sourceReference": 0
}
```
must be converted back to:
```json
"source": {
  "name": "list.scala",
  "path": "/home/piquerez/scalacenter/expression-evaluator-test/.metals/readonly/dependencies/cats-core_2.12-2.3.0-sources.jar/cats/instances/list.scala",
  "sourceReference": 0
}
```


tgodzik:
Great work! I will take a look next week. Does it cover the cases from the Metals BreakpointSuite?

adpi2:
@tgodzik I have copied the `ClassBreakppointSuite` from Metals.

I had to change the expected result of the 3 Scala 3 tests but that's because it is more accurate now.

```diff
test("method-scala3") {
      check(
        """|package a.b
          |def method() = {
          |>>  println(0)
          |}
          |""".stripMargin,
-         "a.b.Main$package",
+        "a.b.Main$package$",
        scalaVersion = ScalaVersion.`3`
      )
    }
    

    test("inner-class-scala3"){
      check(
        """|package a
          |
          |@main 
          |def helloWorld(): Unit = {
          |  object Even {
          |>>  def unapply(s: String): Boolean = s.size % 2 == 0
          |  }
          |}
          |
          |""".stripMargin,
-        "a.Main$package",
+        "a.Main$package$Even$1$",
        scalaVersion = ScalaVersion.`3`
      )
    }
    
    test("optional-braces") {
      check(
        """|package a
          |
          |@main 
          |def hello(): Unit = 
          |  greet("Alice")
          |  greet("Bob")
          |  System.exit(0)
          |
          |def greet(name: String) = 
          |  val message = s"Hello, $name!"
          |>>println(message)
          |
          |""".stripMargin,
-         "a.Main$package",
+         "a.Main$package$",
        scalaVersion = ScalaVersion.`3`
      )
    }
```

core/src/main/scala/ch/epfl/scala/debugadapter/internal/ClassPathEntryLookUp.scala
```diff
@@ -0,0 +1,263 @@
...
+      case SourceDirectory(directory) =>
+        getAllSourceFiles(entry, FileSystems.getDefault, directory).toSeq
+      case StandaloneSourceFile(absolutePath, relativePath) =>
+        Seq(SourceFile(entry, relativePath, absolutePath.toUri))
+    }
+  }
+
+  private def getAllSourceFiles(entry: SourceEntry, fileSystem: FileSystem, root: Path): Iterator[SourceFile] = {
+    if (Files.exists(root)) {
+      val sourceMatcher = fileSystem.getPathMatcher("glob:**.{scala,java}")
```
tgodzik, line 184:
```suggestion
      val sourceMatcher = fileSystem.getPathMatcher("glob:**.{scala,java,sc}")
```
do we want to future proof it if we ever implement run for Ammonite?

core/src/main/scala/ch/epfl/scala/debugadapter/internal/SourceLookUpProvider.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  override def getFullyQualifiedName(uriRepr: String, lines: Array[Int], columns: Array[Int]): Array[String] = {
+    val uri = URI.create(uriRepr)
+    uri.getScheme match {
+      case "dap-fqcn" =>
+        val resolvedName = uri.getSchemeSpecificPart
+        lines.map(_ => resolvedName)
+      case _ =>
+        sourceUriToClassPathEntry.get(uri) match {
+          case None => lines.map(_ => null)
```
tgodzik, line 38: Why map all the lines to null? Is it expected by DAP when not found?

tgodzik (approved):
LGTM! Looks pretty amazing, I just added some smaller comments, but most likely you can ignore them

core/src/main/scala/ch/epfl/scala/debugadapter/internal/ClassPathEntryLookUp.scala
```diff
@@ -0,0 +1,263 @@
...
+
+private case class ClassFile(fullyQualifiedName: String, sourceName: Option[String], relativePath: String) {
+  def className: String = fullyQualifiedName.split('.').last
+  def fullPackage: String = fullyQualifiedName.stripSuffix(className).stripSuffix(".")
+}
+
+private case class SourceFile(entry: SourceEntry, relativePath: String, uri: URI) {
+  def extension: String = relativePath.split('.').last
+  def baseName: String = name.stripSuffix(extension)
+  def name: String = relativePath.split('/').last
```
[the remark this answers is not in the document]
adpi2: It seems to work already because all the tests are passing on windows.

core/src/main/scala/ch/epfl/scala/debugadapter/internal/ClassPathEntryLookUp.scala
adpi2: That's a good idea, never thought of it before. It might require a bit of work though to add some tests. So I'll create an issue and I'll work on it after this PR is merged.

core/src/main/scala/ch/epfl/scala/debugadapter/internal/SourceLookUpProvider.scala
adpi2: It's probably not expected by DAP but it is expected by the java-debug-adapter server which calls this method. Otherwise it throws an `IndexOutOfBoundExpection`, if I remember correctly.

core/src/main/scala/ch/epfl/scala/debugadapter/internal/ClassPathEntryLookUp.scala
tgodzik: It might be enough to just keep an issue open, so that once we want to implement run/debug it's easy to find.

adpi2: https://github.com/scalacenter/scala-debug-adapter/issues/67

PR #68, opened 2021-08-17
Look up into jdk source

adpi2:


core/src/main/scala/ch/epfl/scala/debugadapter/ClassPathEntry.scala
```diff
@@ -1,8 +1,34 @@
...
   def isJar: Boolean = absolutePath.toString.endsWith(".jar")
 }
+
+object ClassPathEntry {
+  def javaRuntime(javaHome: String): Option[ClassPathEntry] =
+    javaRuntime(Paths.get(javaHome))
+
+  /**
+    * From Java 9 onward this method returns null
+    * That's because the rt.jar does not exist anymore
```
er1c, line 18: IIRC, I think the sbt script launcher does some hijinks to copy the rt.jar and make it available.  Let me see if I can dig that up.

er1c: This is what I was thinking of: https://github.com/sbt/java9-rt-export

adpi2: That's helping a lot! Thanks!

core/src/main/scala/ch/epfl/scala/debugadapter/JavaRuntime.scala
```diff
@@ -0,0 +1,59 @@
...
+    } yield {
+      val otherJars = Seq("jre/lib/charsets.jar", "lib/charsets.jar")
+        .map(javaHome.resolve)
+        .filter(Files.exists(_))
+      Java8(Seq(runtimeJar) ++ otherJars, srcZip)
+    }
+  }
+
+  private def java9OrAbove(javaHome: Path, srcZip: Path): Option[JavaRuntime] = {
+    Some("lib/jrt-fs.jar")
```
er1c, line 54: this might need to be semi-dynamically generated or cached on the first use depending on the JRE/JDK your running.  The sbt-launcher does this in https://github.com/sbt/sbt/blob/develop/sbt#L439-L459 

It sticks it inside a version-prefixed path:

```
$ pwd
/home/eric/.sbt/1.0/java9-rt-ext-adoptopenjdk_11_0_11
$ ls -al
total 65652
drwxrwxr-x 2 eric eric     4096 Aug 18 16:12 .
drwxrwxr-x 9 eric eric     4096 Aug 18 16:12 ..
-rw-rw-r-- 1 eric eric 67218802 Aug 18 16:12 rt.jar
```

er1c: Oh SBT also sets the property: `-Dscala.ext.dirs=` so maybe this lib could either support it directly or some other environment variable?

adpi2: The fact that sbt generates a `rt.jar` file is kind of hack to compile Scala on Java 9. Probably on newer Scala version it is not or won't be needed.

So yes I could use the `rt.jar` generated by sbt but I can also use the `lib/jrt-fs.jar` directly to read the `"jrt:/"` filesystem directly. That's kind of the official solution described [here](https://openjdk.java.net/jeps/220):

>A built-in NIO FileSystem provider for the jrt URL scheme ensures that development tools can enumerate and read the class and resource files in a run-time image by loading the FileSystem named by the URL jrt:/, as follows:
>
>FileSystem fs = FileSystems.getFileSystem(URI.create("jrt:/"));
>byte[] jlo = Files.readAllBytes(fs.getPath("modules", "java.base", "java/lang/Object.class"));
>
>The top-level modules directory in this filesystem contains one subdirectory for each module in the image. The top-level packages directory contains one subdirectory for each package in the image, and that subdirectory contains a symbolic link to the subdirectory for the module that defines that package.
>
>For tools that support the development of code for JDK 9 but which themselves run on JDK 8, a copy of this filesystem provider suitable for use on JDK 8 is placed in the lib directory of JDK 9 run-time images, in a file named jrt-fs.jar.

Also that is what `java9-rt-export` does under the hood.

er1c: Sweet, thanks for the explanation!

PR #69, opened 2021-08-18
Expression evaluator

tdudzik:
# Introduction 

This PR introduces a basic implementation of expression evaluation. It's intended to be incrementally improved in the next pull requests. This implementation supports expression evaluation only for Scala 2.12.12.

# Implementation

The expression evaluator was implemented using Java Debug Interface (JDI) and Scala compiler that was extended with two additional phases (`insexpr` and `genexpr`).

When debugging a file like this:

```scala
class Foo {
  def bar(): Int = {
    val a = 1
    val b = 2
    println(s"a + b = ${a + b}")
  }
}
```

with a breakpoint set at the line 5 and expression: `a * b`, evaluator compiles the file and in the phase `insexpr`, just after `parser` phase, inserts the expression at the line of the breakpoint and inserts the expression class in the same package as the expression. So after this phase, the file looks like this:

```scala
class Foo {
  def bar(): Int = {
    val a = 1
    val b = 2
    a * b
    println(s"a + b = ${a + b}")
  }
} 

class Expression {
  def evaluate(names: Array[Any], values: Array[Any]) = {
    val valuesByName = names.map(_.asInstanceOf[String]).zip(values).toMap
    ()
  }
}
```

Then the compiler proceeds with the execution of the rest of the phases and after `delamdafy` phase, when almost all transformations are done (implicits are resolved, etc.), the compiler executes `genexpr` phase. The `genexpr` extracts transformed expression (from the line `5` in this example), extracts all defs (local variables, fields, arguments, etc.) (from the class `Foo` and the method `bar` in this example), and transforms `Expression` class that was inserted before in the following way:
- for every extracted def, it creates a local variable in the method `evaluate` in the form: `val foo: String = valuesByName("foo").asInstanceOf[String]`,
- inserts extracted expression at the end of the `evaluate` method,
- modifies the return type of the `evaluate` method.

After that, the generated `Expression` class is loaded on the running JVM using JDI and the `evaluate` method is called with actual values of local variables, fields, and arguments.


core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/Evaluator.scala
```diff
@@ -0,0 +1,118 @@
...
+
+    val uri = sourceLookUpProvider.getSourceFileURI(fqcn, sourcePath)
+    val content = sourceLookUpProvider.getSourceContents(uri)
+
+    var error: Option[String] = None
+    val result = for {
+      classLoader <- vm.allClasses().asScala.find(_.classLoader() != null).map(_.classLoader()).flatMap(JdiClassLoader(_, thread))
+      variables = frame.visibleVariables().asScala
+      variableNames <- Some(variables.map(_.name()).map(vm.mirrorOf).toList)
+      variableValues <- Some(variables.map(frame.getValue).flatMap(value => boxIfNeeded(value, classLoader, thread)).toList)
```
adpi2, line 33: Could you replace all of the `foo <- Some(bar)` with `foo = bar`

core/src/main/scala/scala/tools/nsc/EvalGlobal.scala
```diff
@@ -0,0 +1,265 @@
...
+  private var expressionOwners: List[Symbol] = _
+
+  override protected def computeInternalPhases(): Unit = {
+    super.computeInternalPhases()
+
+    addToPhasesSet(new InsExpr, "insexpr")
+    addToPhasesSet(new GenExpr, "genexpr")
+  }
+
+  class InsExpr extends Transform with TypingTransformers {
```
adpi2, line 25: Could you add a comment explaining the purpose of `InsExpr`? Same with `GenExpr`? Also inside `GenExpr`, could you add a comment on `ExpressionExtractor`, `DefExtractor`, `GenExprTransformer`?

adpi2 (commented):
Thanks a lot for this wonderful contribution! Could you add some comments to explain the compiler part of the code.

core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/Evaluator.scala
```diff
@@ -0,0 +1,118 @@
...
+      names = variableNames ++ fieldNames ++ thisObjectName.map(Seq(_)).getOrElse(Seq())
+      values <- Some(variableValues ++ fieldValues ++ thisObject.map(Seq(_)).getOrElse(Seq()))
+      systemClass <- classLoader.loadClass("java.lang.System")
+      classPath <- systemClass
+        .invokeStatic("getProperty", List(vm.mirrorOf("java.class.path")))
+        .map(_.toString)
+        .map(_.drop(1).dropRight(1)) // remove quotation marks
+      expressionCompiler <- Some(ExpressionCompiler(classPath, line, expression, names.map(_.value()).toSet))
+      _ <- Some(println(expressionCompiler.dir))
+      expressionClassPath = s"file://${expressionCompiler.dir.toString}/"
```
adpi2, line 47: Currently the windows test are broken because of the URL. This should fix it: 
```suggestion
      expressionClassPath = expressionCompiler.dir.toUri.toURL.toString
```

core/src/main/scala/scala/tools/nsc/EvalGlobal.scala
```diff
@@ -0,0 +1,265 @@
...
+            atPos(tree.pos)(treeCopy.PackageDef(transformed, transformed.pid, transformed.stats :+ parsedExpressionClass))
+          } else {
+            transformed
+          }
+        case _ =>
+          super.transform(tree)
+      }
+
+      private def parseExpression(expression: String): Tree = {
+        // It's need to be wrapped because it's not possible to parse single expression
```
adpi2, line 69:
```suggestion
        // It needs to be wrapped because it's not possible to parse single expression
```

adpi2 (approved):
Thank you for the explanation! This PR is well tested and documented. IMO it is ready to be merged.

PR #70, opened 2021-08-19
Cross compilation of expression evaluator

tdudzik:
This PR implements a cross-compilation for the expression evaluator. It supports the same Scala versions as Metals.


build.sbt
```diff
@@ -86,3 +91,51 @@ lazy val testAgent = project
...
+      "2.12.14",
+      "2.12.13",
+      "2.12.12",
+      "2.12.11",
+      "2.12.10"
+    ),
+    Compile / unmanagedSourceDirectories ++= multiScalaDirectories(
+      (ThisBuild / baseDirectory).value / "expression-compiler",
+      scalaVersion.value
+    ),
```
adpi2, line 116: `scala-2` and `scala-3` are defined by default. So I think this piece of code is useless.

build.sbt
```diff
@@ -86,3 +91,51 @@ lazy val testAgent = project
...
+  val base = root / "src" / "main"
+  val result = mutable.ListBuffer.empty[File]
+  val partialVersion = CrossVersion.partialVersion(scalaVersion)
+  if (isScala2(partialVersion)) result += base / "scala-2"
+  if (isScala3(partialVersion)) result += base / "scala-3"
+  result.toList
+}
+
+def isScala2(v: Option[(Long, Long)]) = v.exists(_._1 == 2)
+def isScala3(v: Option[(Long, Long)]) = v.exists(_._1 == 3)
```
adpi2, line 141: useless too

expression-compiler/src/main/scala-3/scala/tools/nsc/ExpressionCompiler.scala
```diff
@@ -0,0 +1,19 @@
...
+    expressionDir: Path,
+    expressionClassName: String,
+    valuesByNameIdentName: String,
+    classPath: String,
+    code: String,
+    line: Int,
+    expression: String,
+    defNames: ju.Set[String],
+    errorConsumer: Consumer[String]
+  ): Boolean = false
```
adpi2, line 18: Maybe add some error message:
```scala
errorConsumer.accept("The Scala debugger does not yet support Scala 3")
```

adpi2 (commented):
Left some comments. Otherwise it looks good.

adpi2 (approved):

PR #74, opened 2021-08-23
Fix JDI method invocation in expression evaluator

tdudzik:
Fixes error: `Failed to get variables. Reason: com.sun.jdi.InvalidStackFrameException` which occurs after any expression evaluation.


adpi2:
Nice fix @tdudzik. Do you think it is possible to add a test?

tdudzik:
> Nice fix @tdudzik. Do you think it is possible to add a test?

Yes, I've just added a test.

adpi2 (approved):
Thanks

adpi2:
Actually there is a small problem with the unit tests here. Because the `MainDebuggeeRunner.evaluationClassLoader` is always 2.12.4 but that does not work with the `Scala213DebugTests` and `Scala3DebugTests`. So we should probably resolve the correct expression evaluator in `MainDebuggeeRunner` using Coursier (as we do already to resolve the correct Scala library).

PR #76, opened 2021-08-24
Enable evaluation of variable's `toString` methods

tdudzik:
Enables evaluation of variable's `toString` methods as code evaluation is now supported.


core/src/main/scala/ch/epfl/scala/debugadapter/internal/DebugAdapter.scala
```diff
@@ -15,19 +15,6 @@ import java.util.function.Consumer
...
-   * since code evaluation is not supported.
-   *
-   * Debug adapter, when asked for variables, tries to present them in a readable way,
-   * hence it evaluates the `toString` method for each object providing it.
-   * The adapter is not checking if evaluation is supported, so the whole request
-   * fails if there is at least one variable with custom `toString` in scope.
-   *
-   * See usages of [[com.microsoft.java.debug.core.adapter.variables.VariableDetailUtils.formatDetailsValue()]]
-   */
-  DebugSettings.getCurrent.showToString = false
```
adpi2, line 29: Would be could to set it to true only if `runner.expressionClassLoader` is defined.

adpi2: Or rather if `EvaluationProvider` is not a `NoopEvaluationProvider`

PR #77, opened 2021-08-24
Rework resolution of the Scala expression compiler

adpi2:
- Rename expression-compiler into scala-expression-compiler
- Make `resolveEvaluationClassLoader` a private and anonymous task
- Use `BuildInfo` to resolve the expression-compiler

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -36,7 +36,7 @@ lazy val root = project
 
 lazy val core = project
   .in(file("core"))
-  .enablePlugins(SbtJdiTools)
+  .enablePlugins(SbtJdiTools, BuildInfoPlugin)
   .settings(
     name := "scala-debug-adapter",
     libraryDependencies ++= List(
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -48,6 +48,18 @@ lazy val core = project
       Dependencies.coursier % Test,
       Dependencies.coursierJvm % Test
     ),
+    buildInfoKeys := Seq[BuildInfoKey](
+      BuildInfoKey.action("expressionCompilerOrganization")(
```
tdudzik, line 52: It's a minor thing but maybe it's worth to rename everything to `scalaExpression...` to make it consistent.

tdudzik (approved):

build.sbt
adpi2: `scala-expression-compiler` is a better name in the context of Maven Central because that contains all the artifacts of the Java, Kotlin, Groovy, Scala ecosystems.

In the context of this repository it is obvious that `expressionCompiler` is the `scalaExpressionCompiler`. Similarly `core` is obviously the  `scalaDebugAdapterCore` and `sbtPlugin` is the `sbtDebugAdapterPlugin`... IMO shorter names are better as long as there is no ambiguity.

PR #89, opened 2021-09-02
Add support for lambdas and nested methods in expression evaluation

tdudzik:


adpi2:
@tdudzik It is not clear to me why a lambda in an expression was not working before and why it is now working. Can you explain a bit?

Also, is this PR ready to be reviewed? If not, what is missing?

tdudzik:
> @tdudzik It is not clear to me why a lambda in an expression was not working before and why it is now working. Can you explain a bit?
> 
> Also, is this PR ready to be reviewed? If not, what is missing?

It's not. I found out that it doesn't work when lambda uses private fields, also there are some issues with nested methods. I'm currently working on that. I'll try to describe what was the problem when I'm done.

PR #90, opened 2021-09-03
Refactor Evaluator.scala

kpodsiad:
Every time I thought that I understand how the expression evaluator works it always turned out that I don't 😃 
So I decided to go through code and refactor parts which were hard to understand for me. I hope that now is easier to understand or at least not harder :D


tgodzik (approved):
Looks good, what do you think @tdudzik ?

tdudzik (approved):
Looks good, thanks!

PR #107, opened 2021-09-27
Fix expression evaluation

tdudzik:
- fix expression evaluation on breakpoint with a definition or an assignment,
- revert to the non-interactive compiler to avoid `macro has not been expanded` error.


adpi2 (approved):
LGTM!

PR #114, opened 2021-10-05
Introduce ExpressionEvaluationFailed exception

tdudzik:


core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/package.scala
```diff
@@ -23,15 +24,45 @@ package object evaluator {
...
+  }
+
+  private def extractMessage(
+      invocationException: InvocationException
+  )(thread: ThreadReference): String = {
+    val exception = invocationException.exception()
+    val getMessageMethod = method("toString", exception.referenceType())
+    val message: Safe[Value] =
+      invokeMethod(exception, getMessageMethod, List(), thread)
+    message.getResult.toOption.map(_.toString).getOrElse("")
```
adpi2, line 64: It's not a good idea to call `getResult` inside a safe because it prevents the caller to fallback in case of failure (all previous remote object can now be collected). So `extractMessage` should return a `Safe[String]`.

```scala
private def extractMessage(
      invocationException: InvocationException
  )(thread: ThreadReference): Safe[String] = {
    val exception = invocationException.exception()
    val getMessageMethod = method("toString", exception.referenceType())
    val message: Safe[Value] =
      invokeMethod(exception, getMessageMethod, List(), thread)
      message.map(_.toString).recover(_ => "")
}
```

where `recover` can be defined in `Safe` by:
```scala
  def recover[B >: A](f: PartialFunction[Throwable, B]): Safe[B] = {
    new Safe(result.recover(f), dispose)
  }
```

core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/package.scala
```diff
@@ -23,15 +24,45 @@ package object evaluator {
...
-        ObjectReference.INVOKE_SINGLE_THREADED
+      unwrapInvocationException(
+        objRef.invokeMethod(
+          thread,
+          method,
+          args.asJava,
+          ObjectReference.INVOKE_SINGLE_THREADED
+        )
+      )(
+        thread
```
adpi2, line 37: Now that `extractMessage` return a `Safe[String]`, we need some sort of `flatMap` that handles failures of type `Throwable` instead of a result of type `Value`. In the scala-library this is called `recoverWith`.

```scala
Safe(
  objRef.invokeMethod(
      thread,
      method,
      args.asJava,
      ObjectReference.INVOKE_SINGLE_THREADED
    )
).recoverWith {
  case t: InvocationException =>
    extractMessage(t)(thread).map(throw new ExpressionEvaluationFailed(_))
}
```

where `recoverWith` can be defined in `Safe` by:

```scala
def recoverWith[B >: A](f: PartialFunction[Throwable, Safe[B]]): Safe[B] = {
    result match {
      case Failure(exception) if f.isDefinedAt(exception) => 
        val b = f(exception)
        new Safe(b.result, () => {dispose(); b.dispose()})
      case _ => new Safe(result, dispose)
    }
  }
```

adpi2 (changes requested):
It is not a good idea to call `getResult` inside a `Safe` because then all objects can be collected and we cannot do anything with the `Safe` anymore apart from `getResult`.

In general `getResult` should be called only once at the end of the remote evaluation.

I propose an alternative solution in the comment. Note that the solution I propose is inspired by the API of `Future` and `Try` in the standard library.

PR #124, opened 2021-10-06
Fix evalauting in tail recursive annotated functions

tgodzik:
Previously, when trying to evaluate tail recursive function inside itself the compilation of the expression might fail due to the function no longer being tail recursive. Now, we remove the annotation when generating the evaluation and also do not break when evaluating, which would cause the function to stop again.

I also added pprint for nicer debugging messages, but it's not 100% neccessary.


expression-compiler/src/main/scala-2/scala/tools/nsc/EvalGlobal.scala
```diff
@@ -287,7 +317,7 @@ private[nsc] class EvalGlobal(
       override def transform(tree: Tree): Tree = tree match {
         // Don't transform class different than Expression
         case tree: ClassDef if tree.name.decode != expressionClassName =>
-          tree
+          EmptyTree
```
tgodzik, line 320: We should not generate sources for breakpoint source, which we used to generate the expression.

tgodzik: We load up the source that contains the breakpoint, transform it to get the expression and then use the expression in the evaluate method. We can later remove that transformed source, since it's not needed by anything.

core/src/main/scala/ch/epfl/scala/debugadapter/internal/EvaluationProvider.scala
```diff
@@ -61,6 +64,7 @@ private[internal] class EvaluationProvider(
           case Failure(exception) =>
             future.completeExceptionally(exception)
```
adpi2, line 65: I think we should do `isEvaluating.set(false)` here as well. Otherwise the debugger may not stop at the next breakpoint?

adpi2: Same in the case where `evaluator` is `None`. Or even better we should not call `isEvaluating.set(true)` in that case.

adpi2 (changes requested):
I think that `isEvaluating` should always be set to false after `evaluate`.

Also we may need to set `isEvaluating` to true in  `invokeMethod` to cover the case where the user sets a breakpoint in a `toString`. But it should be tested first to see if it is really necessary.

expression-compiler/src/main/scala-2/scala/tools/nsc/EvalGlobal.scala
tgodzik: That's right.

PR #125, opened 2021-10-06
Add support for private method call in expression evaluation

### LINKED ISSUE #121: Cannot evaluate an expression with a private method call
```scala
class Foo {
  def bar(): String = {
    p("a") // breakpoint
  }

  private def p(a: String) = a
}
          
object EvaluateTest {
    new Foo().bar()
  }
}
```

```
> this.p("foo")
```

should return 
```
"foo"
```
but it returns:

```
Cannot evaluate because of com.sun.jdi.InvocationException: Exception occurred in target VM.
```

tdudzik:
Closes #121


adpi2 (approved):
LGTM!

tdudzik:
@adpi2 should we ignore this one failing test?

PR #130, opened 2021-10-07
Add test for nested method evaluation

tgodzik:
Closes https://github.com/scalacenter/scala-debug-adapter/issues/116

Seems this was already fixed in a previously merged PR


core/src/test/scala/ch/epfl/scala/debugadapter/ExpressionEvaluatorSuite.scala
```diff
@@ -482,6 +482,22 @@ abstract class ExpressionEvaluatorSuite(scalaVersion: ScalaVersion)
         _.exists(_ == "\"foo\"")
       )
     }
+
+    "should evaluate nested method" - {
```
tdudzik, line 486: You can also add a similar test but in a class, not in an object. It's possible that it currently works only in objects.

tgodzik: There seems to be no difference, works also inside a class.

Issue #128, opened 2021-10-07
Cannot evaluate expression when computing default value


adpi2:
```scala
object Main12 {
  def main(args: Array[String]): Unit = {
    foo(3)()
  }

  def foo(x: Int)(
=>  y: Int = x + 1
  ): Unit = {
    println("foo")
  }
}
```

```
> x + 1 
Cannot evaluate because of ch.epfl.scala.debugadapter.internal.evaluator.ExpressionCompilationFailed: .
```

PR #131, opened 2021-10-07
Insert evaluated expression into block within ValDef

tgodzik:
This allows to simplify the code and also fixes https://github.com/scalacenter/scala-debug-adapter/issues/128

Suggested by @tdudzik


adpi2 (approved):
LGTM!

PR #150, opened 2021-11-01
Fix #148: Pass classpath to ExpressionEvaluator

### LINKED ISSUE #148: Cannot evaluate when classpath is too big (> 30k characters)
**Metals:** v0.10.8
**Editor:** vscode
**Build Server:** Bloop v1.4.10-8-8d1cbc4f 

When I try and use the expression evaluator that was recently updated in v0.10.8, it fails for me with:

`Cannot evaluate because of ch.epfl.scala.debugadapter.internal.evaluator.ExpressionCompilationFailed: object scala in compiler mirror not found..`

I was doing some spelunking with @tgodzik, and I threw in a `println` in the test to pull out the `classpath`.
```scala
println(s"Classpath -> ${System.getProperty("java.class.path")}")
``` 
The result of that is `/tmp/jvm-forker-manifest7987195697692046528.jar`

Given Bloop uses a separate classpath and always forks the JVM process for tests, could that be what is causing this not to work for me?

adpi2:
When the classpath is too big and exceeds the maximum size of a command line, Bloop creates a `/tmp/jvm-forker-manifest2949990240171503800.jar` file which contains the classpath in a manifest file. But the Scala compiler does not support this and fails with:  `object scala in compiler mirror not found.`.

To fix this we pass the full classpath explicitly to the expression evaluator.

Tested locally:
```scala
package example

object Main {
    println(s"Classpath -> ${System.getProperty("java.class.path")}")
=>  println("Hello, World")
  }
}
```
Classpath is `/tmp/jvm-forker-manifest5241138918947448719.jar` and expression evaluation does work.

### DIFF
--- a/core/src/main/scala/ch/epfl/scala/debugadapter/internal/EvaluationProvider.scala
+++ b/core/src/main/scala/ch/epfl/scala/debugadapter/internal/EvaluationProvider.scala
@@ -24,7 +24,13 @@ private[internal] object EvaluationProvider {
   ): IEvaluationProvider = {
     val evaluator = runner.evaluationClassLoader
       .flatMap(ExpressionCompiler(_))
-      .map(new ExpressionEvaluator(sourceLookUpProvider, _))
+      .map(
+        new ExpressionEvaluator(
+          runner.classPathEntries,
+          sourceLookUpProvider,
+          _
+        )
+      )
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
LGTM! Thanks!

PR #154, opened 2021-11-02
Use runner's classpath instead of classpath entries

tgodzik:
The default is already to use classpath entries so we should use the actual runner classpath.


adpi2 (approved):
Looks good!

PR #157, opened 2021-11-16
Send event with information about tests to the client

kpodsiad:
>I wonder whether it would be ok to just to use output event with the `data` field containing json with the information from the TestEvent. This would all be done in the same session and would be simple to handle on all possible frontends.
>https://microsoft.github.io/debug-adapter-protocol/specification#Events_Output

_Originally posted by @tgodzik in https://github.com/scalameta/metals-vscode/pull/755#discussion_r749273205_

Implementation of the above idea. 
I had some trouble designing scala classes which will be "just working" with gson, especially I couldn't use sealed hierarchy to express union type `SingleTestResult` which is a union type.


kpodsiad:
@adpi2 is there a reason why you didn't use [sbt.ForkMain](https://github.com/sbt/sbt/blob/develop/testing/agent/src/main/java/sbt/ForkMain.java) and created own [ForkMain](https://github.com/scalacenter/scala-debug-adapter/blob/main/sbt/test-agent/src/main/java/ch/epfl/scala/debugadapter/sbtplugin/internal/ForkMain.java)? As far as I can see from diff only major change is the lack of output stream from the socket which isn't used hence logs are logged to the stdout and there are no events sent to the server. I wonder if I can add handling of the output stream or even fall back to the original `ForkMain`. 
Anyway, I'll be grateful for some additional context on this.

adpi2:
Hey @kpodsiad thanks for pinging. So we created our own `ForkMain` to simplify the code, mostly because we did not need any `sbt.testing.Event` communication that was handled by the private [ForkTests.React](https://github.com/sbt/sbt/blob/aa22c02d84aa9bc61687d7dbd4e727339dfb24c1/main-actions/src/main/scala/sbt/ForkTests.scala#L198) in sbt.

It seems that things have changed and that we now need the `sbt.testing.Event` communication. So yes definitely, we can use the original `ForkMain` class.

sbt/plugin/src/main/scala/sbt/SerializableFingerprints.scala
```diff
@@ -0,0 +1,18 @@
...
+ */
+object SerializableFingerprints {
+  // Copied from ForkTests.scala in sbt
+  def forkFingerprint(f: Fingerprint): Fingerprint & Serializable =
+    f match {
+      case s: SubclassFingerprint => new ForkMain.SubclassFingerscan(s)
+      case a: AnnotatedFingerprint => new ForkMain.AnnotatedFingerscan(a)
+      case _ => sys.error("Unknown fingerprint type: " + f.getClass)
+    }
+}
```
kpodsiad, line 18: Borrowed from [bloop](https://github.com/scalacenter/bloop/blob/3eab1f8079cfaf228ec68740daadf5854070c89b/frontend/src/main/scala/sbt/SerializableFingerprints.scala)
Unfortunately, it doesn't work and it's currently unused because of
```shell
Exception in thread "Thread-15" java.lang.IllegalAccessError: failed to access class sbt.ForkMain$SubclassFingerscan from class sbt.SerializableFingerprints$ (sbt.ForkMain$SubclassFingerscan is in unnamed module of loader sbt.internal.MetaBuildLoader @3c46e67a; sbt.SerializableFingerprints$ is in unnamed module of loader sbt.internal.PluginManagement$PluginClassLoader @4eb9f2af)
	at sbt.SerializableFingerprints$.forkFingerprint(SerializableFingerprints.scala:14)
	at ch.epfl.scala.debugadapter.sbtplugin.internal.TestSuitesRunner$Acceptor$1$.$anonfun$run$1(SbtDebuggeeRunner.scala:109)
	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:286)
	at scala.collection.Iterator.foreach(Iterator.scala:943)
	at scala.collection.Iterator.foreach$(Iterator.scala:943)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1431)
	at scala.collection.IterableLike.foreach(IterableLike.scala:74)
	at scala.collection.IterableLike.foreach$(IterableLike.scala:73)
	at scala.collection.AbstractIterable.foreach(Iterable.scala:56)
	at scala.collection.TraversableLike.map(TraversableLike.scala:286)
	at scala.collection.TraversableLike.map$(TraversableLike.scala:279)
	at scala.collection.AbstractTraversable.map(Traversable.scala:108)
	at ch.epfl.scala.debugadapter.sbtplugin.internal.TestSuitesRunner$Acceptor$1$.run(SbtDebuggeeRunner.scala:108)
```
Try to solve this or stick with current, working solution? wdyt @adpi2 @tgodzik

kpodsiad (commented):
Still needs a few touches, but it's a first working version so I wanted to share it with you. I had a lot of trouble with forking fingerprints in `TestSuitesRunner`, I wanted to do it like it's done in bloop but it didn't work so I ended with the solution similar to the former one.

tgodzik (commented):
Looks good! I have some questions and would love it if @adpi2 could take a look.

sbt/plugin/src/main/scala/ch/epfl/scala/debugadapter/sbtplugin/internal/SbtDebuggeeRunner.scala
```diff
@@ -45,18 +55,56 @@ private[debugadapter] final class TestSuitesRunner(
...
+
   override def run(listener: DebuggeeListener): CancelableFuture[Unit] = {
+    val eventHandler = new TestSuiteResultsEventHandler(listener)
+
+    @annotation.tailrec
+    def receiveLogs(is: ObjectInputStream, os: ObjectOutputStream): Unit = {
+      is.readObject() match {
+        case Array(ForkTags.Error, s: String) =>
+          eventHandler.handle(TestSuiteEvent.Error(s))
+          System.out.println("[ERROR] " + s)
```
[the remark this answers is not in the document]
kpodsiad: I did that because [original implementation](https://github.com/sbt/sbt/blob/b2f0b7dadad0f66f2c2511b6541b1bc158fc3a57/testing/agent/src/main/java/sbt/ForkMain.java#L234-L280) logs to the `OutputStream`  while current [was using std out](https://github.com/scalacenter/scala-debug-adapter/blob/989106082776015999c2c6f631d415a3544ce108/sbt/test-agent/src/main/java/ch/epfl/scala/debugadapter/sbtplugin/internal/ForkMain.java#L87-L133)
I think that it should behave like previously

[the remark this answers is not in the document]
kpodsiad: Ok, something went wrong, I'm not sure if it is necessary to have that. I'll have to take a closer look at this

core/src/main/scala/ch/epfl/scala/debugadapter/util/TestUtils.scala
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  def printThrowable(opt: OptionalThrowable): Option[String] = {
+    if (opt.isEmpty) None
+    else Some(stripTestFrameworkSpecificInformation(opt.get().getMessage))
+  }
+
+  private val specs2Prefix = "java.lang.Exception: "
+  private val utestPrefix = "utest.AssertionError: "
+  private val scalaTestPrefix = "org.scalatest.exceptions.TestFailedException: "
```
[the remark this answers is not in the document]
kpodsiad: I've just extracted it from Bloop in order to preserve similar behavior. Not sure if those messages need to be stripped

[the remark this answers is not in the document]
kpodsiad: I've just extracted it from Bloop in order to preserve similar behavior. Not sure if those messages need to be stripped

sbt/plugin/src/main/scala/ch/epfl/scala/debugadapter/sbtplugin/internal/SbtDebuggeeRunner.scala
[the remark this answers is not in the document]
kpodsiad: Ok, now I see how it was working. Previously, the forked process was printing to stdout/stderr and it was forwarded to the debuggee:
```shell
[ERROR] Test a.JunitTestSuite.test1 failed: assertion failed, took 0.002 sec
[ERROR] Test a.JunitTestSuite.test2 failed: assertion failed: should equal to 2, took 0.001 sec
```
Now, the forked process is sending `sbt.ForkTags` & msg through the socket we need to handle those messages by forwarding them to the debuggee listener. I'll borrow logic from the bloop which creates a nice summary for executed tests so the above example will (just like from bloop)
```shell
Test a.JunitTestSuite.test1 failed: assertion failed, took 0.113 sec
Test a.JunitTestSuite.test2 failed: assertion failed: should equal to 2, took 0.002 sec
Execution took 0.12s
4 tests, 2 passed, 2 failed

===============================================
```

tgodzik (approved):
Looks good! I think we should be good to merge it. Are we able to remove some of the logic in Bloop thanks to including them here?

kpodsiad:
> Are we able to remove some of the logic in Bloop thanks to including them here?

Yes, I manage to remove a little bit from Bloop (mostly definitions and utils)

kpodsiad:
What do you think about it @adpi2 ?

core/src/main/scala/ch/epfl/scala/debugadapter/testing/TestSuiteSummary.scala
```diff
@@ -0,0 +1,39 @@
+package ch.epfl.scala.debugadapter.testing
+
+final case class TestSuiteSummary(
+    suiteName: String,
+    duration: Long,
+    tests: java.util.List[SingleTestSummary]
+)
```
adpi2, line 7: This object, and also `TestResultEvent` are sent to the client using the microsoft `ProtocolServer.sendEvent`. Is there a reason why the microsoft `java-debug-core` library does not already contain them. Should we update the version of this library?

core/src/main/java/ch/epfl/scala/debugadapter/SubclassFingerscan.java
```diff
@@ -11,12 +11,12 @@
 
 import java.io.Serializable;
 
-final class SubclassFingerscan implements SubclassFingerprint, Serializable {
+public final class SubclassFingerscan implements SubclassFingerprint, Serializable {
```
adpi2, line 14: I am not sure this class and `AnnotatedFingerscan` should be in the `core` module. Would it be possible to move them to the `sbtPlugin` module?

Maybe the reason you moved them is to use them in Bloop. I am not sure it's a good idea because at one point we will want to use the debug server in Metals directly rather than in sbt or Bloop. But Bloop will still need to be able to run the tests.

adpi2: In the long run it would be good if sbt makes them public. We should definitely create an issue about that if it does not yet exist.

adpi2 (commented):
This PR looks good to me. But I would like to discuss the two points below before we merge.

core/src/main/java/ch/epfl/scala/debugadapter/SubclassFingerscan.java
kpodsiad: You guessed correctly what was the reason, I'll move them to `sbtPlugin` and create an issue in sbt.

Do you know why we need to fork those fingerprints at all? I tried to find out it and I failed, but this still bothers me, why is this necessary.

core/src/main/scala/ch/epfl/scala/debugadapter/testing/TestSuiteSummary.scala
kpodsiad: Although [spec](https://microsoft.github.io/debug-adapter-protocol/specification) allows arbitrary event type `event: string` and content `body?: any` [java-debug-core](https://github.com/microsoft/java-debug/blob/719eb8c3efd3098b75e91c5b5f866c3b1c984ce2/com.microsoft.java.debug.core/src/main/java/com/microsoft/java/debug/core/protocol/Events.java) offers only few concrete event types. Code is unchanged since `0.18` so updating the library version wouldn't solve mentioned problem.

kpodsiad: But I think their approach is quite OK because one can always create own event extending some base class, just like I did with `TestResultEvent`

adpi2: Ok thanks for the explanation.

core/src/main/java/ch/epfl/scala/debugadapter/SubclassFingerscan.java
adpi2: I don't know either why they are needed. The only reason I can think of is that they extends `Serializable` whereas `SubclassFingerprint` and `SubclassFingerscan` does not.

adpi2 (approved):
Thanks @kpodsiad. Great work!

PR #186, opened 2022-02-04
Run even if some source jars are corrupted

tgodzik:
It seems if we are unable to read a source jar, not debug session will be started currently. This can happen if we get a broken source jar such as the one in https://github.com/scalameta/metals/issues/3584

Tested it manually and it seems to help.

### DIFF
--- a/core/src/main/scala/ch/epfl/scala/debugadapter/ClassSystem.scala
+++ b/core/src/main/scala/ch/epfl/scala/debugadapter/ClassSystem.scala
@@ -8,24 +8,26 @@ import java.nio.file.Files
 import java.nio.file.Path
 
 sealed trait ClassSystem {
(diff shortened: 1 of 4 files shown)


adpi2 (changes requested):
Looks good to me. But could you add a test on this broken jar in: https://github.com/scalacenter/scala-debug-adapter/blob/main/core/src/test/scala/ch/epfl/scala/debugadapter/internal/ClassEntryLookUpSpec.scala

tgodzik:
> Looks good to me. But could you add a test on this broken jar in: https://github.com/scalacenter/scala-debug-adapter/blob/main/core/src/test/scala/ch/epfl/scala/debugadapter/internal/ClassEntryLookUpSpec.scala

Added!

adpi2 (approved):
Thanks!

PR #192, opened 2022-02-11
Include non-default classifiers in debug classpath

tgodzik:
Related to https://github.com/scalameta/metals/issues/3623


sbt/plugin/src/sbt-test/debug-session/test-native-lib/src/test/scala/example/BugSpec.scala
```diff
@@ -0,0 +1,17 @@
+package example
+
+import org.scalatest.flatspec.AnyFlatSpec
+import java.io.File
+
+class BugSpec extends AnyFlatSpec {
```
tgodzik, line 6: I wasn't able to actually reproduce the issue in scripted, not sure why. @adpi2 any idea what might be going on?

adpi2: I am not sure what you are trying to do. Here your scripted test check that the debug session can start. It does not check that the test in the debug session passes.

Now if you want to test what's happening in the debug session, you need to connect a debug client and to check the logs that it receives from the debug server:

See for instance, on how to connect a debug client:
https://github.com/scalacenter/scala-debug-adapter/blob/6a6e522b001bec84b82a41aff6ea255d1815eba3/sbt/plugin/src/sbt-test/debug-session/test-suite/build.sbt#L9-L41

tgodzik: I changed the tests to use breakpoints, the second breakpoint should not be stopped at without my fix, but it actually does :/

I feels like we aren't actually testing things properly, but I honestly don't want to put more effort into this, since it's a really small issue.

tgodzik: Och, I actually it fails on CI with my fixes :(

tgodzik: Och, the test doesn't work on JDK 17 :/

I would need to do something to make it work, but I really don't want to delve into that rabit hole :sweat: 

Would be ok to merge without the added test? It's a very niche situation, but I belive it may happen to people.

adpi2: No problem. I deactivated the test on JDK 17.

Also I found why the test was passing without the fix. That was because the exception was thrown just after the second breakpoint (on the same line actually).

adpi2 (approved):
Thanks for the fix

sbt/plugin/src/sbt-test/debug-session/test-native-lib/src/test/scala/example/BugSpec.scala
tgodzik: Och, I was sure it wouldn't get to that second breakpoint :fearful: Thanks!

PR #198, opened 2022-03-02
feat: scala test selection request

kpodsiad:
PR in bsp repository: build-server-protocol/build-server-protocol/pull/304

### DIFF
--- a/sbt/plugin/src/main/contraband/dap.contra
+++ b/sbt/plugin/src/main/contraband/dap.contra
@@ -32,3 +32,23 @@ type ScalaMainClass {
   ## Additional environment variables for the application.
   environmentVariables: [String]
 }
+
+type ScalaTestSuites {
+  ## Test classes selected to be run.
+  suites: [ch.epfl.scala.debugadapter.sbtplugin.internal.ScalaTestSuiteSelection]
+
+  ##  Additional jvmOptions which will be passed to the forked JVM.
+  jvmOptions: [String]
+
+  ## Environment variables should be an array of strings in format KEY=VALUE
+  environmentVariables: [String]
+}
+
+type ScalaTestSuiteSelection {
+  ## Fully qualified name of the test suite class
+  className: String!
+
+  ## List of tests which should be run within this test suite.
+  ## Empty collection means that all of them are supposed to be executed.
+  tests: [String]
+}
--- a/sbt/plugin/src/main/scala/ch/epfl/scala/debugadapter/sbtplugin/DebugAdapterPlugin.scala
+++ b/sbt/plugin/src/main/scala/ch/epfl/scala/debugadapter/sbtplugin/DebugAdapterPlugin.scala
@@ -11,7 +11,8 @@ import sbt.internal.server.ServerHandler
 import sbt.internal.server.ServerIntent
 import sbt.internal.util.complete.Parser
 import sbt.internal.util.complete.Parsers
-import sjsonnew.BasicJsonProtocol
+import sbt.testing.Selector
+import sbt.testing.TestSelector
 import sjsonnew.shaded.scalajson.ast.unsafe.JValue
 import sjsonnew.support.scalajson.unsafe.CompactPrinter
 import sjsonnew.support.scalajson.unsafe.Converter
@@ -31,6 +32,7 @@ object DebugAdapterPlugin extends sbt.AutoPlugin {
   private object DataKind {
     final val ScalaMainClass: String = "scala-main-class"
     final val ScalaTestSuites: String = "scala-test-suites"
+    final val ScalaTestSuitesSelection: String = "scala-test-suites-selection"
     final val ScalaAttachRemote: String = "scala-attach-remote"
   }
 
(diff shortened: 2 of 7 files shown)


sbt/plugin/src/main/contraband/dap.contra
```diff
@@ -32,3 +32,23 @@ type ScalaMainClass {
...
+
+type ScalaTestSuites {
+  ## Test classes selected to be run.
+  suites: [ch.epfl.scala.debugadapter.sbtplugin.internal.ScalaTestSuiteSelection]
+
+  ##  Additional jvmOptions which will be passed to the forked JVM.
+  jvmOptions: [String]
+
+  ## Enviroment variables should be an array of strings in format KEY=VALUE
+  environmentVariables: [String]
```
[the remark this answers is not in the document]
adpi2: I don't think contraband supports `Map`.

Technically, you could define a new record and use it to define `ScalaTestSuites`

```
type EnvironmentVariable {
  variable: String
  value: String
}

type ScalaTestSuites {
  ...  
  environmentVariables: [EnvironmentVariable]
}
```

But this would not be consistent with the already defined `environmentVariables: [String]` in `ScalaMainClass`.
So I would suggest to leave it as it is.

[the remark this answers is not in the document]
kpodsiad: Ok, then I'll leave this as it is

sbt/plugin/src/main/scala/ch/epfl/scala/debugadapter/sbtplugin/DebugAdapterPlugin.scala
```diff
@@ -325,6 +390,38 @@ object DebugAdapterPlugin extends sbt.AutoPlugin {
...
+    }
+
+    val additionalEnv = environmentVariables
+      .foldLeft(Map.empty[String, String]) { case (env, line) =>
+        val idx = line.indexOf("=")
+        if (idx > 0 && idx != line.length - 1) {
+          val key = line.substring(0, idx)
+          val value = line.substring(idx + 1)
+          env + (key -> value)
+        } else env
```
adpi2, line 415: Maybe this would be slightly easier to read:

```suggestion
        line.split('=') match {
           case Array(key, value) => env + (key -> value)
           case _ => env
         }
`````

**Review by adpi2 — COMMENTED** [review]
Thanks @kpodsiad I appreciate your work on this topic. It's great!

It would be great if you could add a scripted test on tests selection and environment variables.

To do so you can find some inspiration from the scripted [`debug-session/test-suite`](https://github.com/scalacenter/scala-debug-adapter/tree/main/sbt/plugin/src/sbt-test/debug-session/test-suite) and create a new test `debug-session/test-suite-selection`.

In the `build.sbt` file of the scripted test there is a `checkDebugSession` that starts a debug server and communicates with it to see that all goes well. In your case I don't think you need any breakpoint. You can use ` client.outputed(...)` to check the printed messages. So if you have two tests and you want to run only one, you can check that one test is printed but not the other one. Also you can have the test that except some env var and prints something.

I hope that helps otherwise I am still available.

File: sbt/plugin/src/main/scala/ch/epfl/scala/debugadapter/sbtplugin/DebugAdapterPlugin.scala (hunk 2 above)
Comment (kpodsiad) [reply sbt/plugin/src/main/scala/ch/epfl/scala/debugadapter/sbtplugin/DebugAdapterPlugin.scala:415]: Seems a lot nicer! I copied this code from Bloop and I didn't think about it enough

File: sbt/plugin/src/sbt-test/debug-session/test-selection-suite/src/test/scala/TestSuite.scala [hunk 3]
```diff
@@ -0,0 +1,17 @@
+import org.junit.Test
+
+class TestSuite {
+
+  @Test
+  def test1() {
+    val mem = Runtime.getRuntime().maxMemory()
+    val env = System.getenv("KEY")
+    val out = s"test1_${mem}_$env"
```
kpodsiad, line 9: I tried
```scala
s"""|test1
       |$mem
       |$env""".stripMargin
```
but it didn't work as I expected

sbt/plugin/src/sbt-test/debug-session/test-selection-suite/build.sbt
```diff
@@ -0,0 +1,41 @@
...
+
+  val client = TestDebugClient.connect(uri)
+  try {
+    client.initialize()
+    client.launch()
+
+    client.configurationDone()
+
+    val ignoredTest = client.outputed(_.category == Category.stdout)
+    val outputEvent = client.outputed(_.category == Category.stdout)
```
kpodsiad, line 28: Here, `outputEvent` contained only `test1` but no more events were available to fetch, so I just went with the easiest way - I concatenated string with `_`

kpodsiad:
It seems like failures aren't related to the changes.

adpi2 (approved):
Thanks, I appreciate the contribution!

kpodsiad:
It was a pleasure, thanks for your guidance!

kpodsiad:
Hey @adpi2, would you mind cutting a release with this PR?

PR #209, opened 2022-05-09
Add .jvmopts and switch to setup-java action

tgodzik:


.jvmopts
```diff
@@ -0,0 +1,6 @@
+-Xss4m
+-Xms1G
+-Xmx2G
+-XX:ReservedCodeCacheSize=1024m
+-XX:+TieredCompilation
+-Dfile.encoding=UTF-8
```
adpi2, line 6: What's the purpose of those options?

tgodzik: just copied over the ones in Metals, they seem to work fine here also, but I can pinpoint exactly which are needed in this case.

It seems though there is an issue with the Windows build anyway, will need to take a look at that

.github/workflows/ci.yml
```diff
@@ -9,23 +9,23 @@ jobs:
...
     steps:
     - uses: actions/checkout@v2.3.4
-    - uses: olafurpg/setup-scala@v10
+    - uses: actions/setup-java@v3
       with:
+        distribution: 'temurin'
         java-version: ${{ matrix.jdk }}
     - name: Cross-publish the expression evaluator
-      run: sbt --client + expressionCompiler/publishLocal
+      run: sbt "--client +expressionCompiler/publishLocal"
```
tgodzik, line 28: It seems that I am unable to make `--client` work on windows :thinking: 

Not sure why that is in this case

ckipp01:
It looks like this has been switched to use the coursier setup action. So I'll go ahead and close this.

PR #210, opened 2022-05-13
Improve Scala 3 evaluation: static objects and private fields

adpi2:
- Add more Scala 3 evaluation tests for `@main` annotations and brace-less syntax
- Fix cast in Scala 3 evaluation of idents
- Refactor Scala 2 evaluator (move phases to different files)
- Add failing tests on accessing static objects and their fields
- Fix evaluation of static objects and private fields


core/src/test/scala/ch/epfl/scala/debugadapter/MoreScala3EvaluationTests.scala
```diff
@@ -0,0 +1,237 @@
...
+      assertInMainClass(source, "EvaluateTest")(
+        Breakpoint(12)(
+          ExpressionEvaluation.success("msg(\"Alice\")", "Hello, Alice!"),
+          ExpressionEvaluation.success("msg1(1)", "Hello, 1!"),
+          ExpressionEvaluation.success("msg2(new Foo)", "Hello, foo!")
+        )
+      )
+    }
+
+    "brace-less syntax: evaluate tail-rec function" - {
```
tgodzik, line 220: Btw. could you add a test that has some generic type in it? It seems that evaluator breaks when we get something like `: List[String]`. IF it's not fixed by this PR I can tak a stab at it.

expression-compiler/src/main/scala-2/scala/tools/nsc/EvaluationBridge.scala
```diff
@@ -23,18 +21,17 @@ final class EvaluationBridge {
       timeoutMillis: Long
   ): Boolean = {
     val settings = new Settings
+    // settings.Xprint.value = List("insert-expression", "typer", "generate-expression")
```
tgodzik, line 24: That might be useful for debugging, right? Maybe add something like `// this code can be used to print the tree after each phase of the debugger`

core/src/test/scala/ch/epfl/scala/debugadapter/MoreScala3EvaluationTests.scala
adpi2: I fixed that and added a test in `ScalaEvaluationTests`

core/src/test/scala/ch/epfl/scala/debugadapter/ScalaEvaluationTests.scala
```diff
@@ -0,0 +1,876 @@
...
+          |""".stripMargin
+      val evaluation =
+        ExpressionEvaluation.success("List(1, 2, 3).map(_ * a * b * c).sum", 36)
+      assertInMainClass(source, "EvaluateTest")(
+        Breakpoint(6)(evaluation),
+        Breakpoint(15)(evaluation)
+      )
+    }
+
+    "evaluate call to anonymous function" - {
```
adpi2, line 807: @tgodzik this one tests the evaluation of the generic local value `f`

expression-compiler/src/main/scala-2/scala/tools/nsc/EvaluationBridge.scala
adpi2: Ok I'll add this comment in a future PR

PR #212, opened 2022-05-13
Smart step

jrossett:


core/src/main/scala/ch/epfl/scala/debugadapter/internal/DebugAdapter.scala
```diff
@@ -25,6 +28,41 @@ private[debugadapter] object DebugAdapter {
...
+          println(method.signature())
+
+          // Check wether the signature looks like a lambda
+          if (method.name().contains("%anonfunc%")) {
+            false
+          }
+
+          val className = method.getClass().getCanonicalName()
+
+          // How to access fqcnToClassPathEntry?
```
jrossett, line 46: @adpi2 How can I access fqcnToClassPathEntry to get the classPathEntryLookup for the right name?

core/src/main/scala/ch/epfl/scala/debugadapter/internal/DebugAdapter.scala
```diff
@@ -25,6 +28,41 @@ private[debugadapter] object DebugAdapter {
...
+    DebugSettings.getCurrent().stepFilters.customStepFilter =
+      new CustomStepFilter {
+        override def skip(method: Method): Boolean = {
+          // TODO: Check name, return type, arguments number and types
+          // If sure => skip
+          // Else step in
+          println(method.signature())
+
+          // Check wether the signature looks like a lambda
+          if (method.name().contains("%anonfunc%")) {
```
jrossett, line 40: @adpi2 is it the right way to do it? I can't find the place where it was written in the intellij project and do not remember how to do it

adpi2: You can find the pattern by decompiling the class file of a class that contains a lambda.

1. Install the Scala 2.13.8 compiler using `cs`
```
$ cs install scalac:2.13.8
```
2. Define a class Foo that contains a lambda. Here the lambda is `x => x + 1`.
```scala
// Foo.scala
package example

class Foo {
  def foo(): Unit = {
    List(1, 2).map(x => x +1)
  }
}
```
3. Compile the class using `scalac`
```
$ scalac Foo.scala
```
It will create the `example/Foo.class` class file 
4. Decompile the class file with `javap`
```
$ javap example.Foo
javap  example.Foo 
Compiled from "Foo.scala"
public class example.Foo {
  public void foo();
  public static final int $anonfun$foo$1(int);
  public example.Foo();
}
```

core/src/main/scala/ch/epfl/scala/debugadapter/internal/DebugAdapter.scala
adpi2: `fqcnToClassPathEntry` should stay private in `SourceLookUpProvider`

What you should do is add a method in `SourceLookUpProvider` that takes a `className` and return the `classFile` or the bytes of the class.
```scala
private[debugadapter] final class SourceLookUpProvider(...) {

  def getClassFile(className: String): ClassFile = {
     // here you can use fqcnToClassPathEntry
  }
}

File: core/src/main/scala/ch/epfl/scala/debugadapter/internal/SourceLookUpProvider.scala [hunk 3]
```diff
@@ -53,6 +53,16 @@ private[debugadapter] final class SourceLookUpProvider(
     classPathEntries.flatMap(_.fullyQualifiedNames)
   private[internal] def allOrphanClasses: Iterable[ClassFile] =
     classPathEntries.flatMap(_.orphanClassFiles)
+
+  private[internal] def getClassFile(className: String): Option[String] = {
+    val classEntryLookup = fqcnToClassPathEntry.get(className).get
```
adpi2, line 58:
```suggestion
    val classEntryLookup = fqcnToClassPathEntry(className)
```

adpi2:
```suggestion
    fqcnToClassPathEntry.get(className)
      .flatMap(lookup => lookup.getClassFile(className))
      
   for {
     lookup <- fqcnToClassPathEntry.get(className)
     classFile <- lookup.getClassFile(className)
   } yield classFile
```

PR #263, opened 2022-07-28
Use run / baseDirectory for debugger

### LINKED ISSUE #262: Debugger should use `run / baseDirectory`
Now cwd (or working directory) is determined using generic build `baseDirectory`, which is common for build, run, tests, virtually everything in the project.

According to sbt documentation it is more correct to use `run` or even `Compile / run` version of it: https://www.scala-sbt.org/1.x/docs/Forking.html#Change+working+directory

The change will allow better control over working directory for debug sessions (that it will not cause side effects on other build parts).

arixmkii:
Fixes #262 

Changed to `run / baseDirectory`, but this could be narrowed down even to `Compile / run / baseDirectory` if being the most strict.

The builds, where the most generic `baseDirectory` was changed should continue to work w/o breakage.

Can retarget this to `2.2.x` if this makes any sense.

### DIFF
--- a/sbt/plugin/src/main/scala/ch/epfl/scala/debugadapter/sbtplugin/DebugAdapterPlugin.scala
+++ b/sbt/plugin/src/main/scala/ch/epfl/scala/debugadapter/sbtplugin/DebugAdapterPlugin.scala
@@ -196,7 +196,7 @@ object DebugAdapterPlugin extends sbt.AutoPlugin {
       val target = Keys.bspTargetIdentifier.value
       val scalaVersion = Keys.scalaVersion.value
       val javaHome = Keys.javaHome.value
-      val workingDirectory = Keys.baseDirectory.value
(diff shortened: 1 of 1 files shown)


arixmkii:
I believe `run / baseDirectory` is better, because debug session is utilized for debugging tests as well if I understand this correctly. And if there is no way to distinguish these scopes, better to keep it compatible with both.

tgodzik (approved):
LGTM from my side, but let's wait for @adpi2 since he is much more versed in sbt than me :sweat_smile:

adpi2 (approved):
Thanks @arixmkii 

> Changed to run / baseDirectory, but this could be narrowed down even to Compile / run / baseDirectory if being the most strict.

This is already scoped to `Compile` in line 88 by
```
inConfig(Compile)(runSettings)
```

Issue #251, opened 2022-06-27
Expression evaluation doesn't work with -Xsource:3


tgodzik:
As discussed in https://github.com/scalacenter/bloop/pull/1741 we might need to add `scalacOptions` field along to scalaVersion.

I can take a look at it next week.

adpi2:
Thank you @tgodzik.

One important detail is that each build target can have its own set of `scalacOptions`, so maybe we should add `scalacOptions` into `ClassPathEntry`. For the library dependencies the `scalacOptions` are unknown, which is a problem: we might not be able to evaluate expressions into a source of a dependency. One way to solve this would be for sbt and other build tools to store the scalac options into the pom file when publishing it.

tgodzik:
Or alternatively we could add -Xsource:3 option by default for the libraries?

adpi2:
Yeah sure but `-Xsource:3` is only a part of the problem. We might also need to enable all experimental features or to add  `-Xsource:future`.

tgodzik:
> Yeah sure but `-Xsource:3` is only a part of the problem. We might also need to enable all experimental features or to add `-Xsource:future`.

yeah, things like kind projector might also be problematic, so probably no solution for that unless libraries publish their options in pom.xml or MANIFEST :/

PR #280, opened 2022-09-23
Configure the `ExpressionCompiler` on each `ManagedEntry`

adpi2:
Introduce `DebugTools` and `DebugToolsResolver` to resolve the expression compilers and the step filter base on the modules and libraries of the debuggee.

At most 2 expression compilers are resolved, one for Scala 2 and one for Scala 3
For both Scala 2 and Scala 3 we want to use the most recent known version:
  - the version of the main module
  - or the version of the scala-library/scala3-library in the classpath
  - or the latest version known by the scala-debug-adapter
That's because we want to be able to read the full classpath.

Examples:
 - if there are Scala 3.1 and 3.2 compiled jars in the classpath, we don't want to use the Scala 3.1 compiler
   because it cannot unpickle the 3.2 classes 
  - similarly, only the latest Scala 2.13 version can unpickle the classes of the latest Scala 3 version


adpi2:
Fixes #251

PR #292, opened 2022-10-13
Introduce DebugTestSuite and refactor all tests

adpi2:


adpi2:
The failing `scala3StepFilter / test` should be fixed by #293.

PR #363, opened 2023-01-04
Register scala-debug-adapter_2.13 for publication

adpi2:
@lefou Next version of the scala-debug-adapter will have Scala 2.13 support.


lefou (approved):
I appreciate a Scala 2.13 version of this project. This change looks good to me. What else is needed to move this forward?

PR #366, opened 2023-01-10
fix: use event duration instead of whole suite result's duration when making a summary

kpodsiad:
reported in https://github.com/scalameta/metals-vscode/issues/1296.

- change `testSuiteResult.duration` to `e.duration`
- extract `TestSuiteEventHandler` to separate file, I think it deserves that
- add some random thoughts scaladoc


adpi2 (approved):
Thanks a lot for this fix and the comments!

PR #385, opened 2023-03-09
Add documentation for releasing the full Scala Debug Adapter or a specific version of the Expression Compiler

adpi2:


CONTRIBUTING.md
```diff
@@ -155,3 +155,25 @@ addSbtPlugin("ch.epfl.scala" % "sbt-debug-adapter" % "3.0.2-SNAPSHOT")
...
   - or right-click on a test and run `Debug test`
   - or add a `Scala` debug configuration in the `.vscode/launch.json` file
+
+## Releasing
+
+### Releasing the full scala-debug-adapter
+
+To publish a new release you can go to the [release page](https://github.com/scalacenter/scala-debug-adapter/releases/new), to create a new tag and generate the release notes.
+
+**The release is not triggered automatically.**
```
ckipp01, line 165: How come we don't include something like:

```
on:
  push:
    tags: ["*"]
```

That can detect a tag push, which would automate this? Is there a reason that we don't want to do that here?

ckipp01 (commented):
Looks like all the tests are failing at
```
debug-session/no-expression-compiler
```

CONTRIBUTING.md
adpi2: The reason is that we need to specify a version for the java-debug-core module which is different from the tag. Of course we could set that version in the build.sbt file and push it. But it is too easy to forget about it, losing half an hour trying to understand why the release failed, and pushing a new tag. It seems more convenient to specify the java-debug version in the workflow input.

ckipp01: Gotcha, makes sense 👍🏼

PR #394, opened 2023-04-03
fix: made debugger recover from not found sources

iusildra:


modules/tests/src/main/scala/ch/epfl/scala/debugadapter/testfmk/DebugTestSuite.scala
```diff
@@ -145,11 +153,11 @@ trait DebugTest {
         print(s"> ${event.output}")
         assertion(event.output.trim)
       case DebugStepAssert(_: StepIn, assertion) =>
-        println(s"Stepping in, at ${topFrame.source.name} ${topFrame.line}")
+        extractThenPrintSource(topFrame)
```
adpi2, line 156: Here it now prints "Stepping out, " instead of "Stepping in, ".

I suggest to do something like this:
```scala
println(s"Stepping in, ${formatFrame(topFrame)}")
```
Where `formatFrame` is a new method that would replace `extractThenPrintSource`.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/ClassEntryLookUp.scala
```diff
@@ -174,97 +174,94 @@ private object ClassEntryLookUp {
       sourceFiles: Seq[SourceFile],
       logger: Logger
   ): ClassEntryLookUp = {
-    if (sourceFiles.isEmpty) ClassEntryLookUp.empty(entry, logger)
```
adpi2, line 177: Is the `empty` method still used, or should we remove it?

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/StepFilterTests.scala
```diff
@@ -221,6 +221,9 @@ abstract class StepFilterTests(protected val scalaVersion: ScalaVersion) extends
       Breakpoint(23),
       StepIn.line(10)
     )
+    checks
+    debuggee = TestingDebuggee.mainClassWithoutSources(source, "example.Main", scalaVersion)
+    checks
```
adpi2, line 226: To make it more explicit, I think we should extract in a new test. The name of test should clearly state than we can step into a Java method even if the source files of the JDK are not found.

For instance we could test that we can step into foo and then step into println.
```scala
packag example

object Main {
    foo("world")
  }

  def foo(name: String): Unit = {
    println(foo)
  }
}
```

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/ClassEntryLookUp.scala
iusildra: It's not used anymore, I removed it

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/ScalaDebugTests.scala
```diff
@@ -43,6 +43,24 @@ class Scala3DebugTest extends ScalaDebugTests(ScalaVersion.`3.1+`) {
 }
 
 abstract class ScalaDebugTests(val scalaVersion: ScalaVersion) extends DebugTestSuite {
+  test("should not crash when sources aren't present".only) {
```
adpi2, line 46:
```suggestion
  test("should not crash when sources aren't present") {
```

adpi2 (approved):
Apart from the `.only`, it looks good to me.

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/ScalaDebugTests.scala
iusildra: :sob:

PR #401, opened 2023-04-05
Evaluate expression asynchronously

### LINKED ISSUE #325: Evaluate expression asynchronously
The `IEvaluationProvider` interface is asynchronous, all methods return an instance of `java.util.concurrent.CompletableFuture`. Unfortunately [our implementation](https://github.com/scalacenter/scala-debug-adapter/blob/main/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/EvaluationProvider.scala) is synchronous.

The goal of this issue is to make the implementation of the `EvaluationProvider` asynchronous using `scala.concurrent.Future`.

One test should be added in [ScalaEvaluationTests](https://github.com/scalacenter/scala-debug-adapter/blob/main/modules/tests/src/test/scala/ch/epfl/scala/debugadapter/ScalaEvaluationTests.scala) to check that there is no bad interaction between two or more concurrent evaluations.

iusildra:
Fix #325

### DIFF
--- a/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/DebugAdapter.scala
+++ b/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/DebugAdapter.scala
@@ -14,6 +14,7 @@ import java.util
 import java.util.Collections
 import java.util.concurrent.CompletableFuture
 import java.util.function.Consumer
+import scala.concurrent.ExecutionContext
 
 private[debugadapter] object DebugAdapter {
 
@@ -23,7 +24,9 @@ private[debugadapter] object DebugAdapter {
    */
   DebugSettings.getCurrent.showStaticVariables = true
 
-  def context(debuggee: Debuggee, tools: DebugTools, logger: Logger, config: DebugConfig): IProviderContext = {
+  def context(debuggee: Debuggee, tools: DebugTools, logger: Logger, config: DebugConfig)(implicit
+      ec: ExecutionContext
+  ): IProviderContext = {
     TimeUtils.logTime(logger, "Configured debugger context") {
       val context = new ProviderContext
       val classEntries = debuggee.classEntries
--- a/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/DebugAdapter.scala
+++ b/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/DebugAdapter.scala
@@ -14,7 +14,6 @@ import java.util
 import java.util.Collections
 import java.util.concurrent.CompletableFuture
 import java.util.function.Consumer
-import scala.concurrent.ExecutionContext
 
 private[debugadapter] object DebugAdapter {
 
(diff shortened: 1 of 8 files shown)


modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/EvaluationProvider.scala
```diff
@@ -42,22 +44,32 @@ private[internal] class EvaluationProvider(
...
+    extends IEvaluationProvider {
 
   private var debugContext: IDebugAdapterContext = _
   private val isEvaluating = new AtomicBoolean(false)
 
+  private def fromTryToFuture[T](t: => Try[T]) = {
+    t match {
+      case Success(value) => Future.successful(value)
+      case Failure(e) => Future.failed(e)
+    }
```
adpi2, line 57: Here we want to run the evaluation of `t` asynchronously.

```suggestion
    Future(t.get)
```

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/EvaluationProvider.scala
```diff
@@ -109,7 +123,7 @@ private[internal] class EvaluationProvider(
           case MethodInvocationFailed(msg, exception) => exception
         }
     }
-    completeFuture(invocation.getResult, thread)
+    completeFuture(fromTryToFuture(invocation.getResult), thread)
```
adpi2, line 126: Here we evaluate invocation synchronously and then we wrap it into a future. It would be best to evaluate it asynchronously.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/DebugAdapter.scala
```diff
@@ -23,7 +24,9 @@ private[debugadapter] object DebugAdapter {
    */
   DebugSettings.getCurrent.showStaticVariables = true
 
-  def context(debuggee: Debuggee, tools: DebugTools, logger: Logger, config: DebugConfig): IProviderContext = {
+  def context(debuggee: Debuggee, tools: DebugTools, logger: Logger, config: DebugConfig)(implicit
+      ec: ExecutionContext
+  ): IProviderContext = {
```
adpi2, line 29:
```suggestion
  def context(debuggee: Debuggee, tools: DebugTools, logger: Logger, config: DebugConfig): IProviderContext = {
```

modules/tests/src/main/scala/ch/epfl/scala/debugadapter/testfmk/DebugTestSuite.scala
```diff
@@ -83,15 +85,15 @@ trait DebugTest {
...
+  private def check(client: TestingDebugClient, attach: Option[Int])(steps: DebugStepAssert*): Unit = {
     client.initialize()
     attach match {
       case Some(port) => client.attach("localhost", port)
       case None => client.launch()
     }
 
     val sourceBreakpoints = steps
-      .map(_.step)
+      .collect { case s: SingleStepAssert[_] => s.step }
```
adpi2, line 96: In Scala 3, we prefer the new `[?]` syntax than `[_]`

```suggestion
      .collect { case s: SingleStepAssert[?] => s.step }
```

modules/tests/src/main/scala/ch/epfl/scala/debugadapter/testfmk/TestingDebugClient.scala
```diff
@@ -274,9 +281,10 @@ class AbstractDebugClient(
     new BufferedWriter(new OutputStreamWriter(output, ProtocolEncoding))
   )
 
-  private val sequenceNumber = new AtomicInteger(1)
+  protected val sequenceNumber = new AtomicInteger(1)
```
adpi2, line 284: I think it would be cleaner if we move this back to `private`. The idea would be to not use the `sequenceNumber.getAndIncrement()` in the `createRequest` method but to use it in the `sendRequest` instead.

adpi2 (approved):

PR #405, opened 2023-04-13
Runtime evaluator

iusildra:
Goal: make an evaluator that would evaluate an expression with reflection by using [jdi](https://docs.oracle.com/en/java/javase/18/docs/api/jdk.jdi/module-summary.html) and [scalameta](https://scalameta.org/)

How it should work:

1. Parse the expression into a tree using scalameta
2. Validate the expression by traversing the scalameta tree, and generating a new tree containing references and any relevant information (arguments, qualifier...) of the corresponding element (or only their name and types if it depends on the current frame), that can be used to cache the expression
    * name resolution (local variable, fields, modules, classes...)
    * method resolution (apply calls, non-generic overloads resolution)
    * type checking (works only on non-generic types because of erasure)
3. Evaluate the expression using the tree generated during the validation

Because of some overloads resolution or access to private field during the debug session, one evaluation mode might fail but not the other, so there must be some priorities:
<p align="center">

</p>


adpi2 (commented):
Some comments

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimePrimitiveOp.scala
```diff
@@ -0,0 +1,217 @@
...
+      case Div | Modulo => None
+      case Less => Some(fractional.lt(x, y))
+      case LessOrEqual => Some(fractional.lteq(x, y))
+      case Greater => Some(fractional.gt(x, y))
+      case GreaterOrEqual => Some(fractional.gteq(x, y))
+    }
+
+    result match {
+      case Some(value) => clsLoader.mirrorOfAnyVal(value)
+      case None => Safe(Failure(new IllegalArgumentException("Division by zero")))
```
adpi2, line 95: This should probably be wrapped in a `MethodInvocationException` which means that the exception happened in the debuggee?

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/JdiClassLoader.scala
```diff
@@ -19,21 +19,54 @@ private[internal] class JdiClassLoader(
...
       classObject <- classClass.invokeStatic(
         "forName",
         "(Ljava/lang/String;ZLjava/lang/ClassLoader;)Ljava/lang/Class;",
-        Seq(classNameValue, mirrorOf(true), this)
+        Seq(classNameValue, trueMirror, this)
       )
     } yield classObject.asClass
 
-  def mirrorOf(str: String): Safe[JdiString] =
+  @inline private def loadThenGetType(className: String): Safe[ClassType] =
```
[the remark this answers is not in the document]
iusildra: not anymore :grinning:

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimePrimitiveOp.scala
```diff
@@ -0,0 +1,217 @@
...
+      rhs: JdiValue,
+      loader: JdiClassLoader
+  ): Safe[JdiValue] =
+    for {
+      l <- lhs.toBoolean
+      r <- rhs.toBoolean
+      result <- this match {
+        case And => loader.mirrorOf(l && r)
+        case Or => loader.mirrorOf(l || r)
+        case Not => loader.mirrorOf(!l)
```
adpi2, line 195: What's the value of rhs in this case?

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimePrimitiveOp.scala
```diff
@@ -0,0 +1,217 @@
...
+sealed trait ObjectOp extends RuntimePrimitiveOp {
+  override def typeWith(lhs: Type, rhs: Type): PrimitiveType =
+    lhs.virtualMachine().mirrorOf(true).`type`().asInstanceOf[PrimitiveType]
+
+  override def evaluate(
+      lhs: JdiValue,
+      rhs: JdiValue,
+      loader: JdiClassLoader
+  ): Safe[JdiValue] = this match {
+    case Eq => loader.mirrorOf(lhs.value == rhs.value)
```
adpi2, line 212: Does this work if I do

```scala
case class Foo(x: String)
val foo1 = Foo("a")
val foo2 = Foo("b")

foo1 == foo2
```

adpi2: foo1 and foo2 are not the same instance but they are equal

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/ValidationMonad.scala
```diff
@@ -0,0 +1,92 @@
+package ch.epfl.scala.debugadapter.internal.evaluator
+
+import java.util.NoSuchElementException
+
+sealed abstract class Validation[+A]() {
```
adpi2, line 5: The file should be `Validation.scala` instead of `ValidationMonad.scala`

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/ValidationMonad.scala
```diff
@@ -0,0 +1,92 @@
...
+  def isValid: Boolean
+  def canContinue: Boolean
```
adpi2, line 7: Not sure what it means. Is it used anywhere? Should we remove it?

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/ValidationMonad.scala
```diff
@@ -0,0 +1,92 @@
...
+    case Valid(value) => ev(this.get)
+    case Recoverable(message) => Recoverable(message)
+    case Unrecoverable(e) => Unrecoverable(e)
+  }
+
+  def get: A
+  def getOrElse[B >: A](f: => B): B
+  def orElse[B >: A](f: => Validation[B]): Validation[B]
+
+  def toOption: Option[A]
```
adpi2, line 22: Is it used? Should we remove it?

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/package.scala
```diff
@@ -11,12 +14,47 @@ package object evaluator {
...
   }
 
+  implicit class TrySeq[A](seq: Seq[Try[A]]) {
+    def traverse: Try[Seq[A]] = {
+      seq.foldRight(Try(Seq.empty[A])) { (safeHead, safeTail) =>
+        safeTail.flatMap(tail => safeHead.map(head => head +: tail))
+      }
+    }
+  }
+
```
adpi2, line 24: This can move to `ScalaExtension` if it is not already there

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/internal/JavaRuntimeEvaluatorTests.scala
```diff
@@ -0,0 +1,252 @@
...
+       |    public static String staticMethod() { return "i am static static_friend_foo"; }
+       |  }
+       |}
+       |
+    """.stripMargin
+}
+class Scala212RuntimeEvaluatorJavaTests extends JavaRuntimeEvaluatorTests(ScalaVersion.`2.12`)
+class Scala213RuntimeEvaluatorJavaTests extends JavaRuntimeEvaluatorTests(ScalaVersion.`2.13`)
+class Scala3RuntimeEvaluatorJavaTests extends JavaRuntimeEvaluatorTests(ScalaVersion.`3.0`)
+class Scala31RuntimeEvaluatorJavaTests extends JavaRuntimeEvaluatorTests(ScalaVersion.`3.1+`)
```
adpi2, line 135: Do we really need to test on all Scala versions? I wonder if it would even be possible to have no Scala version at all.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimePrimitiveOp.scala
```diff
@@ -0,0 +1,217 @@
...
+
+  // TODO: 1 - add support for types that inherit from the types above
+  // TODO: 2 - add some type checking (do 1 before)
+  def apply(lhs: RuntimeValidationTree, rhs: RuntimeValidationTree, op: String): Option[RuntimePrimitiveOp] = {
+    (lhs.`type`, rhs.`type`, op) match {
+      case (_, _, "==") => Some(Eq)
+      case (_, _, "!=") => Some(Neq)
+      case (_: ReferenceType, _, _) if !allowedReferenceTypes.contains(lhs.`type`.name()) => None
+      case (_, _: ReferenceType, _) if !allowedReferenceTypes.contains(rhs.`type`.name()) => None
+      case (_, _, "&&") => Some(And)
```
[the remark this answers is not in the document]
iusildra: yes, got to type check it so i also accept potential subtypes of `java.lang.Boolean` (same goes for numeric operations with `java.lang.Number`)

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/internal/RuntimeEvaluatorTests.scala
```diff
@@ -0,0 +1,332 @@
...
+       |    def greet = "Friendly"
+       |  }
+       |}
+       |
+    """.stripMargin
+}
+
+class Scala212RuntimeEvaluatorTests extends RuntimeEvaluatorTests(ScalaVersion.`2.12`)
+class Scala213RuntimeEvaluatorTests extends RuntimeEvaluatorTests(ScalaVersion.`2.13`)
+class Scala3RuntimeEvaluatorTests extends RuntimeEvaluatorTests(ScalaVersion.`3.0`)
```
adpi2, line 130: The Scala31RuntimeEvaluatorTests is enough for Scala 3

```suggestion
```

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluator.scala
```diff
@@ -0,0 +1,475 @@
...
+    case class PreparedCall(qual: Validation[RuntimeValidationTree], name: String)
+
+    def validate(expression: Stat): Validation[RuntimeValidationTree] = {
+      expression match {
+        case value: Term.Name => validateName(value)
+        case _: Term.This => thisTree.map(tree => ThisTree(tree.value))
+        case sup: Term.Super => ???
+        case _: Term.Apply | _: Term.ApplyInfix | _: Term.ApplyUnary => validateMethod(extractCall(expression))
+        case select: Term.Select => validateSelect(select)
+        case newValue: Defn.Val => ???
```
adpi2, line 138: It should return some Recoverable instead of `???`

```suggestion
        case sup: Term.Super => ???
```

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluator.scala
```diff
@@ -0,0 +1,475 @@
...
+  /* -------------------------------------------------------------------------- */
+
+  private object RuntimeValidation {
+    case class Call(fun: Term, argClause: Term.ArgClause)
+    case class PreparedCall(qual: Validation[RuntimeValidationTree], name: String)
+
+    def validate(expression: Stat): Validation[RuntimeValidationTree] = {
+      expression match {
+        case value: Term.Name => validateName(value)
+        case _: Term.This => thisTree.map(tree => ThisTree(tree.value))
```
adpi2, line 134:
```suggestion
        case _: Term.This => thisTree
```

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluator.scala
```diff
@@ -0,0 +1,475 @@
...
+      }
+
+    @inline private def varTreeByName(name: String): Validation[RuntimeEvaluationTree] =
+      Validation.fromOption(frame.variableByName(name)).map(LocalVarTree(_))
+
+    @inline private def fieldTreeByName(
+        of: Validation[RuntimeValidationTree],
+        name: String
+    ): Validation[RuntimeEvaluationTree] =
+      extractElementIfReference(of) { ref =>
```
adpi2, line 163: I tend to prefer to use `map` instead of custom higher order function

```suggestion
      asReference(of).map { ref =>
```

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluator.scala
```diff
@@ -0,0 +1,475 @@
...
+        for {
+          field <- Validation(ref.fieldByName(name))
+          _ = loadClassOnNeed(field)
+        } yield field match {
+          case Module(module) => ModuleTree(module, of.toOption)
```
adpi2, line 168: Since you can evaluate it as field why do you wrap it in a ModuleTree instead?

```scala
class A {
  object B
}

val a1 = new A
val a2 = new A

a1.B // Do I receive the correct instance of B?
```

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluator.scala
```diff
@@ -0,0 +1,475 @@
...
+            searchAllClassesFor(clsName, typeName)
+              .map(ClassTree(_))
+          }
+      }
+
+      of match {
+        case Some(_: TypeTree) =>
+          lookup.flatMap {
+            case cls if cls.`type`.isStatic() => Valid(cls)
+            case cls => Recoverable(s"Cannot access non-static class from class type $cls")
```
adpi2, line 204: Do you have a test on this?

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluationTree.scala
```diff
@@ -0,0 +1,177 @@
...
+    s"""|MethodTree(
+        |${indent}m= $method,
+        |${indent}args= ${args.map(_.toString(depth + 1)).mkString(",\n" + indent)},
+        |${indent}qual= ${qual.toString(depth + 1)}
+        |${indent.dropRight(1)})""".stripMargin
+  }
+}
+case class StaticMethodTree(
+    method: Method,
+    args: Seq[RuntimeValidationTree],
```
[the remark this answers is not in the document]
adpi2: A method can only take values as argument

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/ValidationMonad.scala
iusildra: just forgot to remove it

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/ValidationMonad.scala
iusildra: yep, you'll see it on the next push

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/internal/JavaRuntimeEvaluatorTests.scala
iusildra: some tests were good on scala 3, but not on scala 2 (typically with inner types, there is a field in the outer types in scala3 but not in scala2), and other were good on scala 2.13+, but not on scala 2.12 (it was the `+:` operator)

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/internal/RuntimeEvaluatorPrimitiveOperationTests.scala
```diff
@@ -0,0 +1,304 @@
...
+       |  def char(): Char = 'a'
+       |}
+       |""".stripMargin
+}
+
+class Scala212RuntimeEvaluatorPrimitiveOperationTests
+    extends RuntimeEvaluatorPrimitiveOperationTests(ScalaVersion.`2.12`)
+class Scala213RuntimeEvaluatorPrimitiveOperationTests
+    extends RuntimeEvaluatorPrimitiveOperationTests(ScalaVersion.`2.13`)
+class Scala3RuntimeEvaluatorPrimitiveOperationTests extends RuntimeEvaluatorPrimitiveOperationTests(ScalaVersion.`3.0`)
```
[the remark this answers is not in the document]
iusildra: yay, less tests to run :grin:

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluator.scala
iusildra: I'll do them afterwards

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluator.scala
iusildra: It does (i've just added a check for that)

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluator.scala
iusildra: i have a test that fails when this check is not present

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimePrimitiveOp.scala
iusildra: i changed it a little bit so the value of rhs is "discarded" only for `Not`, but maybe i should do an early check since `Not` is the only one like this

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluationTree.scala
```diff
@@ -0,0 +1,177 @@
...
+) extends RuntimeEvaluationTree {
+  override lazy val `type`: ReferenceType = value.reference.referenceType()
+  override def toString(depth: Int): String = s"ThisTree($value)"
+}
+
+case class ModuleTree(
+    `type`: ClassType,
+    of: Option[RuntimeValidationTree]
+) extends RuntimeEvaluationTree
+    with TypeTree {
```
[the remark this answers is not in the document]
iusildra: yeah, i should rename it. This trait comes with a specialization of the `type` method that allows me access the `instances` method (return a list of the instance of that type) without a cast

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/BypassExpressionCompilerTests.scala
```diff
@@ -302,29 +302,6 @@ class BypassExpressionCompilerTests extends DebugTestSuite {
...
-         |
-         |  def m2(x: () => Boolean): Boolean =
-         |    x()
-         |}
-         |""".stripMargin
-    implicit val debuggee: TestingDebuggee = TestingDebuggee.mainClass(source, "example.Main", scalaVersion)
-    // the evaluator refuses to bypass the compiler because it does not know
-    // if x is a Function0 or a by-name param
-    check(Breakpoint(10), Evaluation.failed("x"), Breakpoint(13), Evaluation.failed("x"))
-  }
```
adpi2, line 326: Maybe we should add a `MixedEvaluationTests` to test cases like this one.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluator.scala
```diff
@@ -0,0 +1,507 @@
...
+    /* -------------------------------------------------------------------------- */
+    private def extractCall(apply: Stat): Call =
+      apply match {
+        case apply: Term.Apply => Call(apply.fun, apply.argClause)
+        case ColonEndingInfix(apply) => Call(Term.Select(apply.argClause.head, apply.op), List(apply.lhs))
+        case apply: Term.ApplyInfix => Call(Term.Select(apply.lhs, apply.op), apply.argClause)
+        case apply: Term.ApplyUnary => Call(Term.Select(apply.arg, Term.Name("unary_" + apply.op)), List.empty)
+      }
+
+    private def validatePrimitiveMethod(
```
adpi2, line 230:
```suggestion
    private def validatePrimitiveOp(
```

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluationTree.scala
```diff
@@ -0,0 +1,175 @@
...
+        |${indent}f= $field,
+        |${indent}on= $on
+        |${indent.dropRight(1)})""".stripMargin
+  }
+}
+
+/* -------------------------------------------------------------------------- */
+/*                                Method trees                                */
+/* -------------------------------------------------------------------------- */
+case class PrimitiveMethodTree(
```
adpi2, line 85:
```suggestion
case class PrimitiveOpTree(
```

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluator.scala
```diff
@@ -0,0 +1,507 @@
...
+    val value = literal match {
+      case Lit.Boolean(value) => vm.mirrorOf(value)
+      case Lit.Byte(value) => vm.mirrorOf(value)
+      case Lit.Char(value) => vm.mirrorOf(value)
+      case Lit.Double(value) => vm.mirrorOf(value.toDouble)
+      case Lit.Float(value) => vm.mirrorOf(value.toFloat)
+      case Lit.Int(value) => vm.mirrorOf(value)
+      case Lit.Long(value) => vm.mirrorOf(value)
+      case Lit.Short(value) => vm.mirrorOf(value)
+      case Lit.String(value) => vm.mirrorOf(value)
```
adpi2, line 352: A string is a reference, it needs to be wrapped in a `Safe`.

You could use the `JdiClassLoader` to implement this method:

```scala
def fromLitToValue(classLoader: JdiClassLoader: literal: Lit): Safe[JdiValue] = {
  literal.value match {
    case v: AnyVal => Safe(classLoader.mirrorOf(v))
    case s: String => classLoader.mirrorOf(s)
    case () => Safe(classLoader.mirrorOfVoid())
  }
}
```

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluator.scala
```diff
@@ -0,0 +1,507 @@
...
+      .asScalaSeq
+      .filter(method => !method.isPrivate && argsMatch(method, args, frame))
+      .toSeq
+      .toValidation(s"Cannot find methods $funName with args types $args")
+      .map(loadClassOnNeed(_, frame))
+
+  /* -------------------------------------------------------------------------- */
+  /*                                Type checker                                */
+  /* -------------------------------------------------------------------------- */
+  def sameType(got: Type, expected: Type, frame: JdiFrame): Boolean = {
```
adpi2, line 400: `isAssignableFrom` is probably a better name.

```suggestion
  def isAssignableFrom(got: Type, expected: Type, frame: JdiFrame): Boolean = {
```

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluator.scala
```diff
@@ -0,0 +1,507 @@
...
+
+  def loadClassOnNeed[T <: TypeComponent](tc: T, frame: JdiFrame): T = {
+    checkClass(tc.`type`)(tc.typeName, frame)
+    tc
+  }
+
+  def searchAllClassesFor(name: String, in: Option[String], frame: JdiFrame): Validation[ClassType] = {
+    def fullName = in match {
+      case Some(value) if value == name => name // name duplication when implicit apply call
+      case Some(value) if value.endsWith("$") => value + name
```
adpi2, line 475:
```scala
class A {
  object B
}

object A {
  object B
}
```
I wonder how the classes of B are encoded. You should add a test maybe.

iusildra: i does not compile :grin:

iusildra:
the last one was '$outer evaluation' and not just evaluation :sweat_smile:

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/EvaluationProvider.scala
```diff
@@ -160,12 +158,22 @@ private[internal] class EvaluationProvider(
...
       Failure(new EvaluationFailed(s"Cannot evaluate '$expression' with $mode mode"))
     }
-  }
+
+  private def prepare(expression: String, frame: JdiFrame): Try[PreparedExpression] =
+    if (mode.allowRuntimeEvaluation)
+      RuntimeEvaluation(frame, logger).validate(expression) match {
+        case Valid(tree) => Success(RuntimeExpression(tree))
+        case MethodCall(tree: RuntimeEvaluationTree) if mode.allowScalaEvaluation =>
+          compilePrepare(expression, frame).orElse(Success(RuntimeExpression(tree)))
```
adpi2:
```suggestion
         case MethodCall(tree: RuntimeEvaluationTree) if mode.allowScalaEvaluation =>
          compilePrepare(expression, frame).orElse(Success(RuntimeExpression(tree)))
        case Valid(tree) => Success(RuntimeExpression(tree))
```

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/EvaluationProvider.scala
```diff
@@ -160,12 +158,22 @@ private[internal] class EvaluationProvider(
...
+        case Fatal(e) => Failure(e)
+        case Recoverable(_) | CompilerRecoverable(_) => compilePrepare(expression, frame)
+      }
+    else compilePrepare(expression, frame)
 
   private def evaluate(expression: PreparedExpression, frame: JdiFrame): Try[Value] = {
     expression match {
       case logMessage: PlainLogMessage => messageLogger.log(logMessage, frame)
-      case localValue: LocalValue => simpleEvaluator.evaluate(localValue, frame)
+      case RuntimeExpression(tree) => RuntimeEvaluation(frame, logger).evaluate(tree).getResult.map(_.value)
```
adpi2, line 176: Should be wrapped into an `evaluationBlock`. This is to prevent the breakpoints to pause the evaluation of an expression.

Maybe the simplest is to wrapped the entire method body in the `evaluationBlock`

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/BypassExpressionCompilerTests.scala
```diff
@@ -5,7 +5,7 @@ import ch.epfl.scala.debugadapter.testfmk.*
 class BypassExpressionCompilerTests extends DebugTestSuite {
   private val scalaVersion = ScalaVersion.`3.1+`
   protected override def defaultConfig: DebugConfig =
-    super.defaultConfig.copy(evaluationMode = DebugConfig.SimpleEvaluationOnly)
+    super.defaultConfig.copy(evaluationMode = DebugConfig.RuntimeEvaluationOnly)
```
adpi2, line 8: Should we move those tests to `RuntimeEvaluatorTests`? Or remove them if they are redundant?

adpi2 (approved):
Good job @iusildra!

PR #417, opened 2023-05-02
Refactor JDI wrappers

adpi2:
Some of the logic in `ScalaEvaluator` is moved to JDI wrappers so that it can be reused more easily. Also the methods in the package object are now encapsulated in the `JdiObject` wrapper.


modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/JdiFrame.scala
```diff
@@ -0,0 +1,41 @@
...
+    def getClassLoaderRecursively(depth: Int): Option[ClassLoaderReference] =
+      if (depth == thread.frameCount) None
+      else {
+        Option(thread.frame(depth).location.method.declaringType.classLoader)
+          .orElse(getClassLoaderRecursively(depth + 1))
+      }
+
+    val classLoader = getClassLoaderRecursively(depth)
+      .getOrElse(throw new Exception("Cannot find any classloader in the stack trace"))
+    JdiClassLoader(classLoader, thread)
```
iusildra, line 24: Maybe only wrap the last lines in the `Safe`

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/JdiObject.scala
```diff
@@ -2,24 +2,52 @@ package ch.epfl.scala.debugadapter.internal.evaluator
...
-  def getFieldValue(name: String): Value = {
+) extends JdiValue(reference, thread) {
+  def getField(name: String): JdiValue = {
     val field = reference.referenceType.fieldByName(name)
-    reference.getValue(field)
+    JdiValue(reference.getValue(field), thread)
+  }
+
+  def invoke(methodName: String, args: Seq[JdiValue]): Safe[JdiValue] = {
+    val m = reference.referenceType.methodsByName(methodName).asScala.head
```
iusildra, line 17: same as above (overloaded methods)

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/JdiObject.scala
```diff
@@ -2,24 +2,52 @@ package ch.epfl.scala.debugadapter.internal.evaluator
...
+      }
   }
+
+  private def extractMessage(invocationException: InvocationException, thread: ThreadReference): Safe[String] =
+    JdiObject(invocationException.exception(), thread)
+      .invoke("toString", List())
+      .map(_.asString.stringValue)
+      .recover { case _ => "" }
+
+  def fields: Seq[(String, JdiValue)] =
```
iusildra, line 48: Why `Seq[(String, JdiValue)]` instead of a `Map[String, JdiValue]` ?

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/JdiValue.scala
```diff
@@ -0,0 +1,57 @@
...
+    }
+  }
+
+  def derefIfRef: JdiValue =
+    value match {
+      case ref: ObjectReference if JdiValue.refTypes.contains(ref.referenceType.name) =>
+        asObject.getField("elem")
+      case _ => this
+    }
+}
```
iusildra, line 30: What is the goal of this method ?

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/ScalaEvaluator.scala
```diff
@@ -113,33 +101,29 @@ private[internal] class ScalaEvaluator(
...
-  ): Safe[(Seq[StringReference], Seq[Value])] = {
-    val frame = frameRef.current()
-    val thisObjectOpt = Option(frame.thisObject) // this object can be null
-    def extractVariablesFromFrame(): Safe[(Seq[StringReference], Seq[Value])] = {
-      val localVariables = frame.visibleVariables().asScala.toSeq.map(v => v.name -> frame.getValue(v))
-      val thisObject = thisObjectOpt.filter(_ => !localVariables.exists(_._1 == "$this")).map("$this".->)
+  ): Safe[(Seq[JdiString], Seq[JdiValue])] = {
+    def extractVariablesFromFrame(): Safe[(Seq[JdiString], Seq[JdiValue])] = {
+      val localVariables = frameRef.variablesAndValues().map { case (variable, value) => (variable.name, value) }
+      val thisObject = frameRef.thisObject().filter(_ => !localVariables.contains("$this")).map("$this".->)
```
iusildra, line 109: What does it do ? (line 109)

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/ScalaEvaluator.scala
```diff
@@ -113,33 +101,29 @@ private[internal] class ScalaEvaluator(
...
-      val fields = thisObject.referenceType.fields.asScala.toList
-      val fieldNames = fields.map(_.name).map(classLoader.mirrorOf).traverse
-      val fieldValues = fields
-        .map(field => thisObject.getValue(field))
-        .map(value => boxIfPrimitive(value, classLoader))
-        .traverse
-      Safe.join(fieldNames, fieldValues)
+    def extractFields(thisObject: JdiObject): Safe[(Seq[JdiString], Seq[JdiValue])] = {
+      val fields = thisObject.fields
+      val names = fields.map(_._1).map(classLoader.mirrorOf).traverse
```
iusildra, line 124: why do you need the `classLoader.mirrorOf` method to extract the fields' name ?

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/ScalaEvaluator.scala
```diff
@@ -162,71 +147,18 @@ private[internal] class ScalaEvaluator(
...
-    val unboxedValues = localVariables()
+  private def updateVariables(variableArray: JdiArray, frame: JdiFrame): Safe[Unit] = {
+    val unboxedValues = frame
+      .variables()
       .zip(variableArray.getValues)
       .map { case (variable, value) =>
-        if (isPrimitive(variable)) unboxIfPrimitive(value, frame.thread)
+        if (variable.`type`.isInstanceOf[PrimitiveType]) value.unboxIfPrimitive
         else Safe(value)
       }
```
iusildra, line 157:
```suggestion
.map {
        case (variable: PrimitiveType, value) => value.unboxIfPrimitive
        case (_, value) => Safe(value)
}
```

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/ScalaEvaluator.scala
```diff
@@ -162,71 +147,18 @@ private[internal] class ScalaEvaluator(
...
-        JdiPrimitive.box(value.value(), classLoader, thread)
-      case value: IntegerValue =>
-        JdiPrimitive.box(value.value(), classLoader, thread)
-      case value: LongValue =>
-        JdiPrimitive.box(value.value(), classLoader, thread)
-      case value: ShortValue =>
-        JdiPrimitive.box(value.value(), classLoader, thread)
-      case value => Safe(value)
-    }
+      yield for ((variable, value) <- frame.variables().zip(values))
```
iusildra, line 161: chained `.map` would be simpler to understand than a nested `for` in the `yield` clause, wdyt ?

iusildra (commented):
Nice refactoring :grinning:
Just a few questions

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/JdiArray.scala
```diff
@@ -1,39 +1,21 @@
...
 
-  def setValues(values: Seq[Value]): Unit = reference.setValues(values.asJava)
+  def setValues(values: Seq[JdiValue]): Unit = reference.setValues(values.map(_.value).asJava)
 
-  def getValues: Seq[Value] = reference.getValues.asScala.toSeq
+  def getValues: Seq[JdiValue] = reference.getValues.asScala.toSeq.map(JdiValue(_, thread))
+}
+
+object JdiArray {
+  def apply(value: Value, thread: ThreadReference): JdiArray =
```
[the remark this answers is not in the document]
adpi2: The value is the array. Maybe I should call it `arrayValue` though.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/JdiClass.scala
```diff
@@ -0,0 +1,77 @@
...
+    cls: ClassType,
+    thread: ThreadReference
+) extends JdiObject(cls.classObject, thread) {
+
+  def initialized: Boolean = cls.isInitialized
+  def className: String = cls.name
+  override def classLoader: JdiClassLoader = JdiClassLoader(cls.classLoader, thread)
+
+  def newInstance(args: Seq[JdiValue]): Safe[JdiObject] = {
+    val ctr = cls.methodsByName("<init>").asScala.head
```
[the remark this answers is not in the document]
adpi2: This method only works reliably if there is no overload. Otherwise it could throw an exception that the number of arguments are incorrect or that they don't have the expected types.

Currently the only way to deal with overloads is to use the other `newInstance` method that takes a `signature` as parameter.

There is no overload resolution based on the types of the passed `args` yet. If we need it we can implement it later.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/JdiClass.scala
```diff
@@ -0,0 +1,77 @@
...
+      }
+    }
+    loadArgumentsRecursively()
+  }
+
+  def getStaticField(fieldName: String): Safe[JdiValue] =
+    Safe(cls.getValue(cls.fieldByName(fieldName))).map(JdiValue(_, thread))
+
+  def invokeStatic(methodName: String, args: Seq[JdiValue]): Safe[JdiValue] = {
+    cls.methodsByName(methodName).forEach(m => println(m.signature))
```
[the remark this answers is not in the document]
adpi2: :hand_over_mouth:

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/JdiFrame.scala
adpi2: :+1:

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/JdiObject.scala
adpi2: In `ScalaEvaluator` we rely on the fact that the variables are always in the same order. I should add a comment.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/JdiValue.scala
adpi2:
```scala
def foo = {
  var x: Int = 1
  def bar = {
    x = 2
  }
}
```

At runtime, `x` is of type `IntRef` so that its value can be updated by the local method `bar`. But when we evaluate `x` we expect an `Int`, so we need to "deref" it.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/ScalaEvaluator.scala
adpi2: To implement `AnyVal` the compiler can create the `$this` local variable that hides the real `$this`.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/ScalaEvaluator.scala
adpi2: We create a `Seq[JdiString]` that we then pass to the expression class (as a `JdiArray`) which lives in the debuggee.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/ScalaEvaluator.scala
adpi2: :+1:

iusildra (approved):
good to go

PR #422, opened 2023-05-10
Skip configuration if noDebug in launchRequest

adpi2:
If `noDebug = true` in the launch request, the debugger skips loadding the `SourceLookUpProvider` (mapping class files to source files), resolving the expression compiler and the step filter. It directly starts the debuggee (main class or test), without enabling any debug feature: no breakpoint, no evaluation. It makes starting the main program or the tests a lot faster.


adpi2:
## with sbt as the build server

With ` noDebug = true`, the DAP launch request is 3s faster (it goes from 4s to 1s).

Unfortunately there is still a lot of time spent on BSP to process all the requests sent by Metals to the client, after clicking the `test` or `debug test` button:
- `buildTarget/compile` ~2s
- `buildTarget/scalaTestClasses` ~4s
- `buildTarget/jvmRunEnvironment` ~3s
- `buildTarget/scalaMainClasses` ~2s
- `buildTarget/startDebugSession` ~4s

Because sbt processes all requests sequentially it takes a total of 15s. To improve this we can try the following:
- make each individual request faster on sbt side
- process requests in parallel in sbt

## With Bloop as the build server

With ` noDebug = true`, the DAP launch request is also 3 seconds faster.

This is a significant improvement because Bloop only takes a few 100 ms to process all the other BSP requests.

So total it takes 1s to start the test instead of 4s.

PR #434, opened 2023-05-26
Module lookup fix

iusildra:
There was a bug with my first implementation. Given the following code
```scala
case class Foo() {
  case class FriendFoo() {
    case class InnerFriendFoo() { val str = "inner friend foo"}
    object InnerFriendFoo { val str = "object inner friend foo"}
  }
  object FriendFoo
  def friendFoo = FriendFoo
}
```
Calling `Foo().friendFoo.InnerFriendFoo.str` could pass the validation even if incorrect

The reason was a bad management of module & nested type name encoding. `Foo().friendFoo` is typed as `Foo$FriendFoo$` and `Foo$FriendFoo$InnerFriendFoo$` is the encoded name of `InnerFriendFoo`. So it passed validation


adpi2 (approved):
LGTM

PR #439, opened 2023-05-31
Pre evaluation

### LINKED ISSUE #438: Pre-evaluation of fields in runtime validation of expression
To get a more precise runtime type and avoid casting, the runtime validation can pre-evaluate any field.

```scala
class Foo

class Bar extends Foo:
  val bar = 42

object Main:
  val foo: Foo = new Bar

  def main(args: Array[String]) =
    println("") // evaluation of foo.bar should be allowed without casting
```

However, if we do so, we would not be able to cache the validation anymore. I suggest we have two modes: one mode that can pre-evaluated but cannot cache (for the debugging console and the watches) and one mode that cannot pre-evaluate but can cache (for the conditional breakpoints, and the logpoints).

iusildra:
Fixes #438


iusildra:
The first commit is mainly restructuration so I can easily switch validators based on the context (debug console, watches...)

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeDefaultValidator.scala
```diff
@@ -0,0 +1,256 @@
...
+import scala.meta.parsers.*
+import Helpers.*
+import RuntimeEvaluatorExtractors.*
+import scala.util.Failure
+import scala.util.Success
+
+case class Call(fun: Term, argClause: Term.ArgClause)
+case class PreparedCall(qual: Validation[RuntimeTree], name: String)
+
+class RuntimeDefaultValidator(val frame: JdiFrame, val logger: Logger) extends RuntimeValidator {
```
adpi2, line 16: There is no other subclass of `RuntimeValidator`, so maybe you should merge `RuntimeDefaultValidator` into `RuntimeValidator`.

iusildra: I'll keep the trait, just in case

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluationHelpers.scala
```diff
@@ -19,7 +19,7 @@ private[evaluator] object Helpers {
     val tpe = classLoader
       .mirrorOfLiteral(literal.value)
       .map(_.value.`type`)
-      .getResult
+      .extract
```
adpi2, line 22: There is a memory leak here. If the lit is a string it will get created in the heap, but will never get garbage collected.
```suggestion
      .getResult
```

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/Safe.scala
```diff
@@ -20,6 +20,9 @@ class Safe[+A] private (
     private val result: Try[A],
     private val dispose: () => Unit
 ) {
+  def extract[B](f: A => B): Try[B] = result.map(f)
+  def extract: Try[A] = extract(identity)
```
adpi2, line 24:
```suggestion
  def extract: Try[A] = result
```

adpi2 (approved):

PR #450, opened 2023-06-08
Method boxing overloads + added private methods

### LINKED ISSUE #445: Fix method lookup when boxing/unboxing occurs
According the [Java specifications](https://docs.oracle.com/javase/specs/jls/se20/html/jls-15.html#jls-15.12.2), the overloads resolution is performed in 2 steps

1. Without boxing/unboxing
2. If no methods were found, then a second round with boxing/unboxing is performed

Currently, there is only one round with unboxing

Todo:
* add a step without unboxing/boxing in `ch.epfl.scala.debugadapter.internal.evaluator.RuntimeEvaluationHelpers.scala#methodsByNameAndArgs`
* add a test in `ch.epfl.scala.debugadapter.internal.RuntimeEvaluatorTests.scala`

iusildra:
Fixes #445 

I also removed the filter on private methods, we might want to be able to call them. Extra caution is required when these methods have side effects though

### DIFF
--- a/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/EvaluationProvider.scala
+++ b/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/EvaluationProvider.scala
@@ -172,7 +172,9 @@ private[internal] class EvaluationProvider(
           compilePrepare(expression, frame).orElse(Success(RuntimeExpression(tree)))
         case Valid(tree) => Success(RuntimeExpression(tree))
         case Fatal(e) => Failure(e)
-        case _: Invalid => compilePrepare(expression, frame)
+        case inv: Invalid =>
+          println(s"\u001b[31mInvalid expression: ${inv.exception}\u001b[0m")
+          compilePrepare(expression, frame)
       }
     else compilePrepare(expression, frame)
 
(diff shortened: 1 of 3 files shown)


modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/EvaluationProvider.scala
```diff
@@ -172,7 +172,9 @@ private[internal] class EvaluationProvider(
           compilePrepare(expression, frame).orElse(Success(RuntimeExpression(tree)))
         case Valid(tree) => Success(RuntimeExpression(tree))
         case Fatal(e) => Failure(e)
-        case _: Invalid => compilePrepare(expression, frame)
+        case inv: Invalid =>
+          println(s"\u001b[31mInvalid expression: ${inv.exception}\u001b[0m")
```
adpi2, line 176: If you want to log something you should use an instance of `ch.epfl.scala.debugadapter.Logger`

adpi2 (approved):

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/EvaluationProvider.scala
iusildra: it's not really for logging purposes, but for testing. Sometimes I wanted to be sure that the validation failed and not the evaluation

PR #456, opened 2023-06-12
Apply method and arrays

### LINKED ISSUE #452: apply() methods with runtime evaluator
### Context
Whenever there is an method call, the runtime evaluator will search in order:
1. For an apply on a **module** or **class** (e.g. `Foo(42)`)
2. For a method of the same name
3. For an apply call shadowed by another method (e.g. `foo()`, where `foo` returns a `Foo$`)
4. Repeat in the outer class, if the `$outer` field is present

### Problem
Other 'implicit' `apply()` methods aren't handled, particularly for functional collections which elements can be accessed with an `apply()` method

```[tasklist]
### Tasks
- [x] Add support for these methods (idea: generalize the current `validateApplyCall` & `validateImplicitApplyCall` methods on `RuntimeDefaultValidator`)
- [x] Translate the `apply` call on `Array` with specific getter
```

iusildra:
Fixes #452

### DIFF
--- a/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/JdiArray.scala
+++ b/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/JdiArray.scala
@@ -14,7 +14,7 @@ class JdiArray(arrayRef: ArrayReference, thread: ThreadReference) extends JdiObj
 
   def getValues: Seq[JdiValue] = arrayRef.getValues.asScala.toSeq.map(JdiValue(_, thread))
 
-  def getValue(i: Int): Safe[JdiValue] = Safe(JdiValue(arrayRef.getValue(i), thread))
+  def getValue(i: Int): JdiValue = JdiValue(arrayRef.getValue(i), thread)
 }
 
 object JdiArray {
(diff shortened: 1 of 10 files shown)


modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeTree.scala
```diff
@@ -125,6 +125,37 @@ object PrimitiveBinaryOpTree {
     }
 }
 
+case class ArrayAccessorTree private (array: RuntimeEvaluableTree, index: RuntimeEvaluableTree, `type`: Type)
```
adpi2, line 128: Maybe `ArrayElemTree` is easier to understand than `ArrayAccessorTree`. We often use `accessor` to designate the generated setter and getter of fields in Scala.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/JdiArray.scala
```diff
@@ -13,6 +13,8 @@ class JdiArray(arrayRef: ArrayReference, thread: ThreadReference) extends JdiObj
   def setValues(values: Seq[JdiValue]): Unit = arrayRef.setValues(values.map(_.value).asJava)
 
   def getValues: Seq[JdiValue] = arrayRef.getValues.asScala.toSeq.map(JdiValue(_, thread))
+
+  def getValue(i: Int): Safe[JdiValue] = Safe(JdiValue(arrayRef.getValue(i), thread))
```
adpi2, line 17: The result does not need to wrapped in a `Safe`:
```suggestion
  def getValue(i: Int): JdiValue = JdiValue(arrayRef.getValue(i), thread)
```

adpi2 (approved):

PR #466, opened 2023-06-21
Pre evaluate outer and Extends new evaluation

iusildra:


modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/Validation.scala
```diff
@@ -81,7 +81,7 @@ object Validation {
       if (value == null) Recoverable("Found null value, expected non-null value")
       else Valid(value)
     } catch {
-      case t: Throwable => Fatal(new Exception(t))
+      case t: Throwable => CompilerRecoverable(new Exception(t))
```
adpi2, line 84: If something unexpected happen, for instance a disconnection, it is probably not recoverable.

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/internal/RuntimeEvaluatorTests.scala
```diff
@@ -616,6 +616,54 @@ abstract class RuntimeEvaluatorTests(val scalaVersion: ScalaVersion) extends Deb
...
+         |    val aAAaaa1 = new aAA.AAA(42)
+         |    val aAAaaa2 = new aAA.AAA(43)
+         |    println("ok")
+         |  }
+         |}
+         |""".stripMargin
+    implicit val debuggee = TestingDebuggee.mainClass(source, "example.Main", scalaVersion)
+    check(
+      Breakpoint(28),
+      Evaluation.success("new a.AA") { res => res.startsWith("A$AA@") },
```
adpi2, line 662: If we stop in `A` (`this` is of type `A`) can we evaluate `new AA` ?

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/internal/RuntimeEvaluatorTests.scala
```diff
@@ -616,6 +616,54 @@ abstract class RuntimeEvaluatorTests(val scalaVersion: ScalaVersion) extends Deb
       Evaluation.success("test(new Integer(1))", "boxed int")
     )
   }
+
+  test("Should instantiate inner classes") {
```
adpi2, line 626: Does it work also with Java classes ?

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/Validation.scala
iusildra: Marked it as an issue for a next PR

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/internal/RuntimeEvaluatorTests.scala
iusildra: Nop :upside_down_face:

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/internal/RuntimeEvaluatorTests.scala
iusildra: Now it should

PR #472, opened 2023-06-27
Evaluate flow control

iusildra:
Allows to evaluate (and eventually pre-evaluate) flow control at runtime. For instance

```scala
case class Test(i: Int)
def test(t: Test) = s"test ${t.i}"

if (true) 1 else 2                    // res0: Int = 1
(if (true) Test(1) else Test(2)).i    // res1: Int = 1
(if (true) Test(1) else 2).i          // res2: Int = 1 thanks to pre-evaluation
test(if(true) Test(1) else 1)         // res3: String = "test 1"
(if (m()) Test(1) else 2).i          // fails because the `if` cannot be pre-evaluated
```

### DIFF
--- a/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeDefaultValidator.scala
+++ b/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeDefaultValidator.scala
@@ -31,12 +31,13 @@ class RuntimeDefaultValidator(val frame: JdiFrame, val logger: Logger) extends R
 
   def validate(expression: Stat): Validation[RuntimeEvaluableTree] =
     expression match {
+      case lit: Lit => validateLiteral(lit)
       case value: Term.Name => validateName(value.value, thisTree)
       case _: Term.This => thisTree
       case sup: Term.Super => Recoverable("Super not (yet) supported at runtime")
       case _: Term.Apply | _: Term.ApplyInfix | _: Term.ApplyUnary => validateMethod(extractCall(expression))
       case select: Term.Select => validateSelect(select)
-      case lit: Lit => validateLiteral(lit)
+      case branch: Term.If => validateIf(branch)
       case instance: Term.New => validateNew(instance)
       case _ => Recoverable("Expression not supported at runtime")
     }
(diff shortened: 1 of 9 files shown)


modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeDefaultValidator.scala
```diff
@@ -252,6 +253,20 @@ class RuntimeDefaultValidator(val frame: JdiFrame, val logger: Logger) extends R
...
+  /*                           Flow control validation                          */
+  /* -------------------------------------------------------------------------- */
+
+  def validateIf(tree: Term.If): Validation[RuntimeEvaluableTree] = {
+    lazy val objType = loadClass("java.lang.Object").extract.get.cls
+    for {
+      cond <- validate(tree.cond)
+      thenp <- validate(tree.thenp)
+      elsep <- validate(tree.elsep)
+      ifTree <- IfTree(cond, thenp, elsep, isAssignableFrom(_, _), objType)
```
adpi2, line 267: The type of the `IfTree` should be the common super type of `thenp` and `elsep`.

```scala
trait A:
  val x: String = "a"

class B extends A
class C extends A

(if (m()) new B else new C).x // should be valid
```

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluationHelpers.scala
```diff
@@ -214,6 +215,43 @@ private[evaluator] class RuntimeEvaluationHelpers(frame: JdiFrame) {
...
+      Some {
+        commonAncestors
+          .reduce { (acc, elem) =>
+            if (acc._1._2 < elem._1._2 && acc._2._2 < elem._2._2) acc
+            else elem
+          }
+          ._1
+          ._1
+      }
+  }
```
adpi2, line 253: If I understand correctly this piece of code, I think you can simplify it like this:

```scala
private def getAncestors(of: Type): Seq[ReferenceType] =
  Iterator.iterate(of)(cls => cls.getSuperClass()).takeWhile(_ != null).toSeq

def extractCommonType(tp1: Type, tpe2: Type): Option[Type] = {
  val ancestors1 = getAncestors(tpe1)
  val ancestors2 = getAncestors(tpe2)
  ancestors1.find(ancestors2.contains)
}
```

adpi2: Also maybe `getAncestors` should be `getSuperClasses` and `extractCommonType` should be `extractCommonSuperClass`

adpi2 (approved):

PR #476, opened 2023-07-03
Fix unexpected static member access

### LINKED ISSUE #470: Name conflicts between variables and methods
Calling a method that has the same name as a variable, with the same arguments as one of the variable's companion object `apply` method leads to evaluator crash. For instance, with the following debuggee:
```scala
object Main {
    val x = 1
    val test = Test(-1)
  }
  def test(x: Int): String = s"int $x"
  def test(t: Test): String = s"test ${t.i}"
  case class Test(i: Int)
}
```
And the expression `test(1)`, the evaluation fails because it finds `Main$Test$.apply(Int)` as a match (calling a static method on an instance, wich can result in unexpected behavior)

That is because during method resolution, we first look for "indirect apply" first, such as `Test(1) <=> Test.apply(1)` or `f(1) <=> f.apply(1)` to avoid conflicts when a class `Foo` is nested with its companion in another class/module (look at `RuntimeDefaultValidator#findMethod` and the test `Should access to multiple layers of nested types`). The reason can be found in the bytecode, this class

```scala
class Test {
  class Inner
  object Inner
}
```
is translated as such
```javap
class Test {
  ...
  public final Test$Inner$ Inner()
  ...
}
```
So when evaluating `Test().Inner()` with reflection, it calls the instance method `Test#Inner()`. Looking for an "indirect apply" first solved this issue, but is blocking here.

An idea would be to consider the call to the `Inner()` method at the debug console level as a constructor call

iusildra:
Fixes #470


modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluatorExtractors.scala
```diff
@@ -25,6 +25,16 @@ protected[internal] object RuntimeEvaluatorExtractors {
       unapply(tree.`type`).map(_ => tree.asInstanceOf[RuntimeEvaluableTree])
   }
 
+  object ModuleCall {
+    def unapply(m: Method): Boolean = {
+      val rt = m.returnTypeName
+      val noArgs = m.argumentTypeNames.size == 0
+      val isSingleton = rt.endsWith("$")
+      val isSingletonInstantiation = rt.stripSuffix("$").endsWith(m.name)
+      noArgs && isSingleton && isSingletonInstantiation
```
adpi2, line 34: For extra safety:

```suggestion
      val isSingleton = rt.endsWith("$" + m.name + "$")
      noArgs && isSingleton
```

PR #481, opened 2023-07-04
Log exceptions

### LINKED ISSUE #480: Log exceptions
## Description
As stated in [this pr review](https://github.com/scalacenter/scala-debug-adapter/pull/466), some exceptions must stop the evaluation and not fallback to the compiler (`VMDisconnectedException`...).
I switched back from returning a `Fatal` on exception to `CompilerRecoverable` after encountering some cases where it blocked the compiler from evaluating valid expressions. However, some exeptions are `Fatal` in any case and should stop the evaluation

## Objectives

```[tasklist]
- [x] Identify 'fatal' exceptions
- [ ] Stop evaluation on fatal ones,
- [ ] Continue evaluation with compiler for others, but log them to be able to correct them later
```

iusildra:
Fixes #480 

For now it simply print in the tests when an exceptions is thrown. Ideally it should be printed by a Logger, but I need to find a way to do so without having to pass it each time a `Validation` is instantiated

### DIFF
--- a/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/Validation.scala
+++ b/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/Validation.scala
@@ -11,8 +11,9 @@ import com.sun.jdi.InvalidStackFrameException
 import com.sun.jdi.AbsentInformationException
 import com.sun.jdi.InvocationException
 import com.sun.jdi.VMOutOfMemoryException
+import ch.epfl.scala.debugadapter.Logger
 
-sealed abstract class Validation[+A] {
+sealed abstract class Validation[+A](implicit logger: Logger) {
   def isValid: Boolean
   def isInvalid: Boolean = !isValid
   def isEmpty: Boolean = isInvalid
@@ -40,7 +41,7 @@ sealed abstract class Validation[+A] {
   def toTry: Try[A]
 }
 
-final case class Valid[+A](value: A) extends Validation[A]() {
+final case class Valid[+A](value: A)(implicit logger: Logger) extends Validation[A]() {
   override val isValid: Boolean = true
   override def map[B](f: A => B): Validation[B] = Validation(f(value))
   override def flatMap[B](f: A => Validation[B]): Validation[B] =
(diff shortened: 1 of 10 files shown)


modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/Validation.scala
```diff
@@ -64,62 +72,73 @@ sealed abstract class Invalid(val exception: Exception) extends Validation[Nothi
...
+    extends Unrecoverable(exception)
 
 object Validation {
-  def apply[A](input: => A): Validation[A] = {
+  private def handler(t: Throwable)(implicit logger: Logger): Invalid = t match {
+    case e @ (_: VMDisconnectedException | _: ObjectCollectedException) => Fatal(e)
+    case e @ (_: InvalidStackFrameException | _: AbsentInformationException) => Fatal(e)
+    case e @ (_: InvocationException | _: VMOutOfMemoryException) => Fatal(e)
+    case e: Exception =>
+      println(s"\u001b[35mUnexpected error while validating: ${e}\u001b[0m")
```
adpi2, line 94: Why don't you use the logger here?

modules/java-debug
adpi2, line 1: Do you know why there is an update of `java-debug` here? It's weird because it says that it contains 0 updated files.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/Validation.scala
iusildra: Forgot to change it

modules/java-debug
iusildra: Yes, it's the merge commit (the previous version was the one on my branch)

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/Validation.scala
```diff
@@ -33,7 +41,7 @@ sealed abstract class Validation[+A] {
   def toTry: Try[A]
 }
 
-final case class Valid[+A](value: A) extends Validation[A]() {
+final case class Valid[+A](value: A)(implicit logger: Logger) extends Validation[A]() {
```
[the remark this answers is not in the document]
iusildra: I thought too, but passing it as implicit params in `map`/`flatMap` reduces the implicit parameters count

adpi2 (approved):
LGTM

PR #482, opened 2023-07-04
Update Java versions on CI

adpi2:
- Updated Java versions on CI
- Added Java 20 in CI
- Check cross-compilation of `Test` in CI
- Fix runtime instanciation of inner class in Java 20

### DIFF
--- a/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeDefaultValidator.scala
+++ b/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeDefaultValidator.scala
@@ -236,13 +236,11 @@ class RuntimeDefaultValidator(val frame: JdiFrame, val logger: Logger) extends R
   def validateNew(newValue: Term.New): Validation[RuntimeEvaluableTree] = {
     val tpe = newValue.init.tpe
     val argClauses = newValue.init.argClauses
-
     for {
       args <- argClauses.flatMap(_.map(validate(_))).traverse
-      outerFqcn = thisTree.toOption
-      (outer, cls) <- validateType(tpe, outerFqcn)(validate)
-      newInstance <- newInstanceTreeByArgs(cls.`type`, args)
-        .orElse(outer.recoverable.flatMap(outer => newInstanceTreeByArgs(cls.`type`, outer +: args)))
+      (outer, cls) <- validateType(tpe, thisTree.toOption)(validate)
+      allArgs = outer.filter(_ => needsOuter(cls.`type`)).toSeq ++ args
+      newInstance <- newInstanceTreeByArgs(cls.`type`, allArgs)
     } yield newInstance
   }
 
--- a/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeDefaultValidator.scala
+++ b/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeDefaultValidator.scala
@@ -240,13 +240,9 @@ class RuntimeDefaultValidator(val frame: JdiFrame, val logger: Logger) extends R
     for {
       args <- argClauses.flatMap(_.map(validate(_))).traverse
       outerFqcn = thisTree.toOption
-      cls <- validateType(tpe, outerFqcn)(validate)
-      qualInjectedArgs = cls._1 match {
-        case Some(q) => q +: args
-        case None => args
-      }
-      method <- methodTreeByNameAndArgs(cls._2, "<init>", qualInjectedArgs, encode = false)
-      newInstance <- NewInstanceTree(method)
+      (outer, cls) <- validateType(tpe, outerFqcn)(validate)
+      newInstance <- newInstanceTreeByArgs(cls.`type`, args)
+        .orElse(outer.recoverable.flatMap(outer => newInstanceTreeByArgs(cls.`type`, outer +: args)))
     } yield newInstance
   }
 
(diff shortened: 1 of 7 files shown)


modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluationHelpers.scala
```diff
@@ -109,25 +107,28 @@ private[evaluator] class RuntimeEvaluationHelpers(frame: JdiFrame) {
...
+      methodsByNameAndArgs(ref, NameTransformer.encode(funName), args.map(_.`type`)).flatMap {
         case ModuleCall() =>
           Recoverable("Accessing a module from its instanciation method is not allowed at console-level")
         case mt => toStaticIfNeeded(mt, args, tree)
       }
     case _ => Recoverable(new IllegalArgumentException(s"Cannot find method $funName on $tree"))
   }
 
+  def newInstanceTreeByArgs(cls: ClassType, args: Seq[RuntimeEvaluableTree]): Validation[NewInstanceTree] =
+    methodsByNameAndArgs(cls, "<init>", args.map(_.`type`))
```
iusildra, line 129:
```suggestion
    methodsByNameAndArgs(cls, "<init>", args.map(_.`type`), encode = false)
```
You must block the encoding, otherwise it will be called as `$lessinit$greater`

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeDefaultValidator.scala
```diff
@@ -240,13 +240,9 @@ class RuntimeDefaultValidator(val frame: JdiFrame, val logger: Logger) extends R
...
-      cls <- validateType(tpe, outerFqcn)(validate)
-      qualInjectedArgs = cls._1 match {
-        case Some(q) => q +: args
-        case None => args
-      }
-      method <- methodTreeByNameAndArgs(cls._2, "<init>", qualInjectedArgs, encode = false)
-      newInstance <- NewInstanceTree(method)
+      (outer, cls) <- validateType(tpe, outerFqcn)(validate)
+      newInstance <- newInstanceTreeByArgs(cls.`type`, args)
+        .orElse(outer.recoverable.flatMap(outer => newInstanceTreeByArgs(cls.`type`, outer +: args)))
```
iusildra, line 245: I think this might lead to the evaluation of weird expressions:
```scala
class Test() {
  class InnerTest()
}
```
Then evaluating `new InnerTest(new Test)` while inside `Test` would succeed. It allows manual 'injection' of the outer class instance. WDYT ?

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/package.scala
```diff
@@ -35,6 +35,11 @@ package object evaluator {
     }
   }
 
+  implicit class OptionExtensions[A](opt: Option[A]) {
+    def recoverable: Validation[A] =
+      opt.map[Validation[A]](Valid.apply).getOrElse(Recoverable("Not found"))
+  }
+
```
iusildra, line 42: There is already a `Validation.fromOption` that does the same thing. Which one do you think is best ? (so I can adapt my other PRs)

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluationHelpers.scala
```diff
@@ -125,6 +124,14 @@ private[evaluator] class RuntimeEvaluationHelpers(frame: JdiFrame) {
     case _ => Recoverable(new IllegalArgumentException(s"Cannot find method $funName on $tree"))
   }
 
+  def needsOuter(cls: ClassType): Boolean =
+    cls
+      .methodsByName("<init>")
+      .asScala
+      .headOption
+      .flatMap(_.argumentTypeNames.asScala.headOption)
+      .exists(argType => cls.name.startsWith(argType + "$"))
```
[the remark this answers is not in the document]
adpi2: If the class needs an outer, then all its constructor would require it as first argument. But maybe I should check all the constructors, for extra safety.

adpi2:
@iusildra it's ready to be reviewed again

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluationHelpers.scala
```diff
@@ -109,25 +106,41 @@ private[evaluator] class RuntimeEvaluationHelpers(frame: JdiFrame) {
...
         case mt => toStaticIfNeeded(mt, args, tree)
       }
     case _ => Recoverable(new IllegalArgumentException(s"Cannot find method $funName on $tree"))
   }
 
+  def needsOuter(cls: ClassType): Boolean =
+    cls
+      .methodsByName("<init>")
+      .asScala
+      .filter(init => init.declaringType.name == cls.name)
```
iusildra, line 131: Are there cases where this predicate is wrong ?

iusildra (approved):
LGTM

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluationHelpers.scala
adpi2: I think `methodsByName` also returns the `<init>` method of upper classes.

PR #484, opened 2023-07-05
Enhance class search

### LINKED ISSUE #503: Non standard class names
As the class / outer class lookup, inner type resolution etc. are implemented, the `$` represents the delimiter between outer/inner classes. Which is not always the case, especially if the class has a 'fancy' name, e.g. `A:B`
I did not see many classes with names like that, but still, it should be handled

iusildra:
Optimize class search with an index
Allow manipulation of classes with special characters in their name (`+`, `&`...)
Allow partially qualified class name (e.g. `mutable.Seq`)

Also fixes #503


modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluationHelpers.scala
```diff
@@ -262,76 +298,67 @@ private[evaluator] class RuntimeEvaluationHelpers(frame: JdiFrame) {
...
-      }
+    def candidates = sourceLookup.classesByName(name)
 
-    finalCandidates
-      .toValidation(s"Cannot find module/class $name, has it been loaded ?")
-      .map { cls => ClassTree(checkClassStatus(cls)(cls.name()).get.asInstanceOf[ClassType]) }
+    def bestMatch = candidates.size match {
+      case 0 | 1 => candidates
+      case _ => candidates.filter(_.startsWith(in.getOrElse(declaringTypeName)))
+    }
```
adpi2, line 331: `candidates` is called twice here. It should probably be defined as a `val` (or a `lazy val`)

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluationHelpers.scala
```diff
@@ -262,76 +298,67 @@ private[evaluator] class RuntimeEvaluationHelpers(frame: JdiFrame) {
...
+      .flatMap { loadClass }
+  }
+
+  def searchClasses(name: String, in: Option[String]): Validation[ClassTree] = {
+    if (isFQCN(name)) loadClass(name)
+    else searchAllClassesFor(name, in)
+  }
+
+  def isFQCN(name: String): Boolean = {
+    val regex = """([\w\.]+)\.([^\.]+)""".r
```
adpi2, line 344: `Foo` is a bad but valid FQCN too.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluationHelpers.scala
```diff
@@ -262,76 +298,67 @@ private[evaluator] class RuntimeEvaluationHelpers(frame: JdiFrame) {
...
+    }
+
+    bestMatch
+      .toValidation(s"Cannot find class $name")
+      .flatMap { loadClass }
+  }
+
+  def searchClasses(name: String, in: Option[String]): Validation[ClassTree] = {
+    if (isFQCN(name)) loadClass(name)
+    else searchAllClassesFor(name, in)
```
adpi2, line 340: Isn't it slower to call `loadClass` instead of `searchAllClassesFor`?

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluationHelpers.scala
```diff
@@ -13,9 +13,20 @@ import scala.util.Try
...
 
 import RuntimeEvaluatorExtractors.*
+import ch.epfl.scala.debugadapter.internal.SourceLookUpProvider
+import scala.collection.mutable.Buffer
 
-private[evaluator] class RuntimeEvaluationHelpers(frame: JdiFrame) {
+private[evaluator] class RuntimeEvaluationHelpers(frame: JdiFrame, sourceLookup: SourceLookUpProvider) {
   import RuntimeEvaluationHelpers.*
+  private implicit class IterableExtensions[A](iter: Iterable[A]) {
+    def toValidation(message: String): Validation[A] =
```
adpi2, line 22: The name `toValidation` does not really indicate that we are expecting a single value. Maybe `validateSingle` is a better name.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluationHelpers.scala
iusildra: woops, yes

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluationHelpers.scala
iusildra: Okay, I though classes with the default package could not be imported

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/SourceEntrySearchTree.scala
```diff
@@ -0,0 +1,29 @@
...
+trait ClassSearch {
+  def insert(fqcn: String): Unit
+  def find(className: String): List[String]
+}
+
+class SourceEntrySearchMap extends ClassSearch {
+  private val classes = Map[String, List[String]]().withDefaultValue(List.empty)
+
+  private def getLastInnerType(className: String): Option[String] = {
+    val pattern = """(.+\$)(.+)$""".r
```
adpi2, line 14: What if it does not end with a `$`?

```suggestion
    val pattern = """(.+\$)(.+)$?""".r
```

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/SourceLookUpProvider.scala
```diff
@@ -13,6 +13,9 @@ private[debugadapter] final class SourceLookUpProvider(
     sourceUriToClassPathEntry: Map[URI, ClassEntryLookUp],
     fqcnToClassPathEntry: Map[String, ClassEntryLookUp]
 ) extends ISourceLookUpProvider {
+  private val classSearch = new SourceEntrySearchMap
+  for (entries <- classPathEntries; fqcn <- entries.fullyQualifiedNames) classSearch.insert(fqcn)
```
adpi2, line 17: I would prefer to use an immutable data structure:

```suggestion
val classSearch: Map[String, Seq[String]] = classPathEntries
  .flatMap(_.fullyQualifiedNames)
  .groupBy(getInnerType)
```

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/SourceLookUpProvider.scala
```diff
@@ -62,6 +65,8 @@ private[debugadapter] final class SourceLookUpProvider(
     classPathEntries.flatMap(_.fullyQualifiedNames)
   private[internal] def allOrphanClasses: Iterable[ClassFile] =
     classPathEntries.flatMap(_.orphanClassFiles)
+  private[internal] def classesByName(name: String) =
```
adpi2, line 68: I think there is a problem because of name encoding.

If a class name contains an operator symbol, e.g. `Foo:Bar`, it will be encoded as `Foo$colonBar`. `classesByName` would receive `Foo:Bar` but `classSearch` would contain `colonBar`. You can add a test to reproduce this bug.

Either you should `decode` all class names before constructing the `classSearch`, or you encode the name and extract the last part before searching in `classSearch`.

The `NameTransformer` class can `encode` and `decode` the class names:
https://github.com/scalacenter/scala-debug-adapter/blob/cb35a4d71927e1a5b7fd3fd1efa37904f4040424/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/NameTransformer.scala#L52

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluationHelpers.scala
adpi2: Yes but you have no guarantee that the `name` is really a fully qualified name. For instance `mutable.Seq` is not a fully qualified name. This seems like a false optimization.

adpi2 (commented):
It would be easier I think not to stack the PR one on top of the other if there are not directly related.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/SourceEntrySearchTree.scala
iusildra: It's not checking if it ends with a `$` but if it ends with the previous group. I could remove it since i take all characters

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeEvaluationHelpers.scala
iusildra: It's not an optimization, it's because when we type `new java.lang.Integer(42)`, without this check, it won't find the class, but I get your point
I'll try to change the map keys to be able to find by package

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/SourceLookUpProvider.scala
iusildra: I should have better read the last line :sob: 
Spent hours trying to find a regex allowing me to extract the last inner type :joy:

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/SourceEntrySearchTree.scala
adpi2: Aaah that makes sense. It's a regex :slightly_smiling_face:

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/SourceLookUpProvider.scala
```diff
@@ -86,6 +90,13 @@ private[debugadapter] object SourceLookUpProvider {
   def empty: SourceLookUpProvider =
     new SourceLookUpProvider(Seq.empty, Map.empty, Map.empty)
 
+  def getClassName(className: String): String = {
```
adpi2, line 93: It's not really clear what this method does. Why do you need to `decode` and then `encode` again?

In the `classSearch` I think we should index by decoded name: `Foo+` instead of `Foo$plus`. That's what the user will type.

Also the name of the method is not really clear. It could be something like `getScalaClassName` instead.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/Validation.scala
```diff
@@ -78,6 +78,7 @@ final case class Recoverable(override val exception: Exception) extends Invalid(
 
 sealed abstract class Unrecoverable(override val exception: Exception) extends Invalid(exception) {
   override def orElse[B >: Nothing](f: => Validation[B]): Validation[B] = this
+  override def transform[B](f: Validation[Nothing] => Validation[B]): Validation[B] = this
```
adpi2, line 81: Do you use this anywhere?

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/internal/RuntimeEvaluatorTests.scala
```diff
@@ -818,6 +818,43 @@ abstract class RuntimeEvaluatorTests(val scalaVersion: ScalaVersion) extends Deb
       Evaluation.success("f1.bar { 1+1; 2+2 }", "bar 4")
     )
   }
+
+  test("Should support names with special characters".only) {
```
adpi2, line 822: only :stuck_out_tongue_closed_eyes:

iusildra: :cry:

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/Validation.scala
iusildra: used to use it at 2 places, and now in the pre-evaluation to avoid unneeded instantiations

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/SourceLookUpProvider.scala
iusildra: it extracts the "clean" class name, without packages or outer classes
good point for the index, it's much simpler :sweat_smile:

iusildra:
I have absolutely no idea of why in Scala 2 there are no breakpoints inside `A+B$$amp$amp$#x` in file `RuntimeEvaluatorTests.scala:898`
I launched the same program in a debug session and placed breakpoints on the whole method. Never reached them

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/SourceLookUpProvider.scala
```diff
@@ -65,8 +69,8 @@ private[debugadapter] final class SourceLookUpProvider(
     classPathEntries.flatMap(_.fullyQualifiedNames)
   private[internal] def allOrphanClasses: Iterable[ClassFile] =
     classPathEntries.flatMap(_.orphanClassFiles)
-  private[internal] def classesByName(name: String) =
-    classSearch.find(name)
+  private[internal] def classesByName(name: String): Seq[String] =
+    classSearch.find(_._1 == name).map(_._2).getOrElse(Seq.empty)
```
adpi2, line 73: `classSearch` is a `Map` so you should do:

```suggestion
    classSearch.get(name).getOrElse(Seq.empty)
```

adpi2 (commented):
One last comment :)

adpi2 (approved):

PR #493, opened 2023-07-07
format operator like types

aymanelamyaghri:


adpi2 (approved):
Some minor comments but otherwise it looks good to me.

modules/unpickler/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3Unpickler.scala
```diff
@@ -100,9 +100,14 @@ class Scala3Unpickler(
       case t: AppliedType if isTuple(t.tycon) =>
         val types = t.args.map(formatType).mkString(",")
         s"(${types})"
-      case t: AppliedType if isAndOrOr(t.tycon) =>
-        val tycon = formatType(t.tycon)
-        t.args.map(formatType).mkString(s" $tycon ")
+      case t: AppliedType if isOperatorLike(t.tycon) =>
```
adpi2, line 103: Actually this only work if the `t.args == 2`

adpi2 (changes requested):

PR #494, opened 2023-07-10
Fix #491: Differentiate getter and method with same name

### LINKED ISSUE #491: Exception in call stack
When debugging `dotty.tools.dotc.rewrites.Rewrites`, the thread cannot be paused because of an exception in the call stack
```
Found 2 matching symbols for dotty.tools.dotc.parsing.Parsers$Parser.expr(dotty.tools.dotc.parsing.Parsers$Location): symbol[Parser.expr] symbol[Parser.expr]
```

adpi2:
- Differentiate getter and method with same name
- Catch all exceptions in `ScalaUnpickler`


adpi2:
@aymanelamyaghri you can review these changes if you'd like

aymanelamyaghri (approved):
LGTM

PR #495, opened 2023-07-10
Multiple outers

### LINKED ISSUE #399: Evaluation involving outer when there are multiple outer
I did not try but I suspect this evaluation to fail

```scala
class A(x: String) {
  class AInner {
    def xx: String = {
      x
    }
  }
}

class B(x: String, y: String) extends A(x) {
  class BInner extends AInner {
    def yy: String = {
      y // evaluate x or y
    }
  }
}
```

Because class `BInner` has two outers, one of type `A` and another of type `B`.

iusildra:
Fixes #399

### DIFF
--- a/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/EvaluationProvider.scala
+++ b/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/EvaluationProvider.scala
@@ -60,7 +60,7 @@ private[internal] class EvaluationProvider(
   override def evaluate(expression: String, thread: ThreadReference, depth: Int): CompletableFuture[Value] = {
     val frame = JdiFrame(thread, depth)
     val evaluator = RuntimeDefaultEvaluator(frame, logger)
-    val validator = RuntimePreEvaluationValidator(frame, logger, evaluator)
+    val validator = RuntimePreEvaluationValidator(frame, evaluator, sourceLookUp, logger)
     val evaluation = for {
       preparedExpression <- prepare(expression, frame, validator)
       evaluation <- evaluate(preparedExpression, frame)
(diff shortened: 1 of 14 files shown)


modules/tests/src/main/scala/ch/epfl/scala/debugadapter/testfmk/TestingLoggers.scala
```diff
@@ -14,9 +14,9 @@ object NoopLogger extends Logger {
...
  */
 object PrintLogger extends Logger {
-  override def debug(msg: => String): Unit = println(s"[DEBUG] $msg")
-  override def info(msg: => String): Unit = println(s"[INFO] $msg")
-  override def warn(msg: => String): Unit = println(s"[WARN] $msg")
-  override def error(msg: => String): Unit = System.err.println(s"[ERROR] $msg")
+  override def debug(msg: => String): Unit = println(s"\u001b[33m[DEBUG] ${msg}\u001b[0m")
+  override def info(msg: => String): Unit = println(s"\u001b[36m[INFO] ${msg}\u001b[0m")
+  override def warn(msg: => String): Unit = println(s"\u001b[91m[WARN] ${msg}\u001b[0m")
+  override def error(msg: => String): Unit = System.err.println(s"\u001b[31m[ERROR] ${msg}\u001b[0m")
```
adpi2, line 20: I would prefer if you use the available `Console.RED` and `Console.RESET` constants instead of the ANSI codes.

iusildra: Oops

iusildra:
It should now be easier to fix #496

adpi2 (approved):
LGTM

PR #502, opened 2023-07-11
find-and-format-local-methods

aymanelamyaghri:


modules/unpickler/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3Unpickler.scala
```diff
@@ -238,8 +283,16 @@ class Scala3Unpickler(
...
       case _ => NameTransformer.encode(symbolName)
-    if method.isExtensionMethod then encodedScalaName == expectedName.stripSuffix("$extension")
-    else encodedScalaName == expectedName
+    matchesLocalMethodOrLazyVal(method) match
+      case None => {
+        if method.isExtensionMethod then encodedScalaName == expectedName.stripSuffix("$extension")
+        else encodedScalaName == expectedName
+      }
+      case Some((methodName, depth)) => {
+        if method.isExtensionMethod then encodedScalaName == expectedName.stripSuffix("$extension")
```
adpi2, line 292: We should add a test of a local method inside a value class to test this.

A value class is a class that extends `AnyVal`. You can probably find one in `Scala3UnpicklerTest` and add a local method to it. Or you can add a new test with something like:

```scala
class Foo(foo: Int) extends AnyVal:
  def m1(bar: Int): Int =
    def m2(x: Int, y: Int): Int = x * y
    m2(foo, bar)
```

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/ScalaStackTraceTests.scala
```diff
@@ -184,4 +184,79 @@ class ScalaStackTraceTests extends DebugTestSuite {
...
+          "Main.main.m.m1: Int",
+          "Main.main.m(t: Int): Int",
+          "Main.main(args: Array[String]): Unit"
+        )
+      )
+    )
+
+  }
+
+  test("should show the correct stack trace  with a local method inside constructor") {
```
adpi2, line 222: There is no such think in Scala as a local method inside a constructor. Here it is just a local method in a normal method that is called in a constructor. I don't think this test is really useful.

adpi2 (commented):
This is looking pretty good already. I left some comments to improve the readability of the code and the test coverage.

adpi2:
You also have some tests that fail:
```
java.lang.IndexOutOfBoundsException: 0
0
	at scala.collection.LinearSeqOps.apply(LinearSeq.scala:131)
	at scala.collection.LinearSeqOps.apply$(LinearSeq.scala:128)
	at scala.collection.immutable.List.apply(List.scala:79)
	at ch.epfl.scala.debugadapter.internal.stacktrace.Scala3Unpickler.findSymbol(Scala3Unpickler.scala:81)
	at ch.epfl.scala.debugadapter.internal.stacktrace.Scala3Unpickler.formatMethod(Scala3Unpickler.scala:49)
	at sun.reflect.GeneratedMethodAccessor18.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at ch.epfl.scala.debugadapter.internal.stacktrace.Scala3UnpicklerBridge.formatScala(Scala3Unpickler.scala:34)
	at ch.epfl.scala.debugadapter.internal.stacktrace.ScalaUnpickler.format(ScalaUnpickler.scala:61)
	at ch.epfl.scala.debugadapter.internal.StackTraceProvider.formatMethod(StackTraceProvider.scala:22)
	at com.microsoft.java.debug.core.adapter.handler.StackTraceRequestHandler.handle(StackTraceRequestHandler.java:82)
	at com.microsoft.java.debug.core.adapter.DebugAdapter.lambda$dispatchRequest$0(DebugAdapter.java:89)
	at java.util.concurrent.CompletableFuture.uniComposeStage(CompletableFuture.java:995)
	at java.util.concurrent.CompletableFuture.thenCompose(CompletableFuture.java:2137)
	at com.microsoft.java.debug.core.adapter.DebugAdapter.dispatchRequest(DebugAdapter.java:88)
	at com.microsoft.java.debug.core.adapter.ProtocolServer.dispatchRequest(ProtocolServer.java:126)
	at ch.epfl.scala.debugadapter.internal.DebugSession.dispatchRequest(DebugSession.scala:207)
	at com.microsoft.java.debug.core.protocol.AbstractProtocolServer.lambda$new$0(AbstractProtocolServer.java:79)
	at io.reactivex.internal.observers.LambdaObserver.onNext(LambdaObserver.java:63)
	at io.reactivex.internal.operators.observable.ObservableObserveOn$ObserveOnObserver.drainNormal(ObservableObserveOn.java:201)
	at io.reactivex.internal.operators.observable.ObservableObserveOn$ObserveOnObserver.run(ObservableObserveOn.java:255)
	at io.reactivex.internal.schedulers.ScheduledRunnable.run(ScheduledRunnable.java:66)
	at io.reactivex.internal.schedulers.ScheduledRunnable.call(ScheduledRunnable.java:57)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:750)
[error] ==> X ch.epfl.scala.debugadapter.Scala30EvaluationTests.evaluate local method in value class  2.406s java.util.NoSuchElementException: next on empty iterator
```

modules/unpickler/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3Unpickler.scala
```diff
@@ -57,17 +59,42 @@ class Scala3Unpickler(
...
+          case Some((methodName, index)) => findLocalMethodOrLazyVal(declaringType,methodName,index)
+          case _ => 
+              val matchingSymbols = declaringType.declarations
+              .collect { case sym: TermSymbol if sym.isTerm => sym }
+              .filter(matchSymbol(method, _))
+               if matchingSymbols.size > 1 then
+               val message =s"Found ${matchingSymbols.size} matching symbols for $method:" +
+               matchingSymbols.mkString("\n")
+               throw new Exception(message)
+               else matchingSymbols.headOption
```
adpi2, line 72: In the CI, the `Check Formatting` failed with a `Parse Error`:

```
org.scalafmt.dynamic.exceptions.PositionExceptionImpl: /home/runner/work/scala-debug-adapter/scala-debug-adapter/modules/unpickler/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3Unpickler.scala:72: error: [dialect scala3] ; expected but else found
               else matchingSymbols.headOption
               ^: /home/runner/work/scala-debug-adapter/scala-debug-adapter/modules/unpickler/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3Unpickler.scala
```

Could you indent the `if else` correctly and then run `scalafmt` to format everything else?

```suggestion
              val matchingSymbols = declaringType.declarations
                .collect { case sym: TermSymbol if sym.isTerm => sym }
                .filter(matchSymbol(method, _))
               if matchingSymbols.size > 1 then
                 val message =s"Found ${matchingSymbols.size} matching symbols for $method:" +
                 matchingSymbols.mkString("\n")
                 throw new Exception(message)
               else matchingSymbols.headOption
```

modules/unpickler/src/test/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3UnpicklerTests.scala
```diff
@@ -202,6 +202,31 @@ abstract class Scala3UnpicklerTests(val scalaVersion: ScalaVersion) extends FunS
...
+         |object Main {
+         |  def main(args: Array[String]): Unit = {
+         |    val a: A = new A("x")
+         |    println(a.m())
+         |  }
+         |}
+         |""".stripMargin
+    val debuggee = TestingDebuggee.mainClass(source, "example.Main", scalaVersion)
+    val unpickler = getUnpickler(debuggee)
+    unpickler.assertFormat("example.A$", "java.lang.String m1$1(java.lang.String t)", "A.m.m1(t: String): String")
```
adpi2, line 227: That's great!

adpi2:
I pushed some additionnal `scalafmt` rules to automate formatting a bit more.

adpi2 (approved):
LGTM

modules/unpickler/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3Unpickler.scala
```diff
@@ -57,17 +59,39 @@ class Scala3Unpickler(
...
+              val message = s"Found ${matchingSymbols.size} matching symbols for $method:" +
+                matchingSymbols.mkString("\n")
+              throw new Exception(message)
+            else matchingSymbols.headOption
+
+  def findLocalMethodOrLazyVal(declaringType: DeclaringSymbol, name: String, index: Int): Option[TermSymbol] =
+    val matchingSymbols =
+      val declaringtpe = declaringType.owner match
+        case s: DeclaringSymbol => s
+        case _ => declaringType
```
adpi2, line 78: Why do you need to take the owner of the `declaringType`. If `declaringType` is `example.A`, `declaringtpe` would be `example` and so you will traverse all the trees of the package `example` which is a lot too much.

modules/unpickler/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3Unpickler.scala
```diff
@@ -57,17 +59,39 @@ class Scala3Unpickler(
...
+                tree match
+                  case DefDef(_, _, _, _, symbol) =>
+                    List((symbol, depth(declaringType, symbol)))
+                  case ValDef(_, _, _, symbol) => List((symbol, depth(declaringType, symbol)))
+                  case _ => List()
+              )((l1, l2) => l1 ++ l2, List())
+            case None => List()
+        )
+        .filter((symbol, depth) => matchTargetName(name, symbol) && depth >= 1)
+    Some(matchingSymbols.sortBy((_, depth) => depth).map((symbol, _) => symbol)(index - 1))
```
adpi2, line 94: The index is not the depth of the symbol but its order of appearance in the `declaringSymbol`. I don't think you need to compute the `depth` at all.

```suggestion
    Some(matchingSymbols(index - 1))
```

See, in my test, there is `m$1`,  `m$2` and `m$3`.

modules/unpickler/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3Unpickler.scala
```diff
@@ -231,6 +256,22 @@ class Scala3Unpickler(
...
+      case pattern(stringPart, numberPart) if (!stringPart.endsWith("$lzyINIT1") && !stringPart.endsWith("$default")) =>
+        Some((stringPart, numberPart.toInt))
+      case _ => None
+
+  private def depth(declaringSymbol: Symbol, symbol: Symbol): Int =
+    symbol.owner match
+      case s: Symbol =>
+        if s.name == declaringSymbol.name then 0
+        else 1 + depth(declaringSymbol, s)
+      case _ => 0
```
adpi2, line 273: Probably not needed

adpi2 (changes requested):
Actually I tested it on the following test and it did not work as expected:

```scala
test("local methods with same name") {
    val source =
      """|package example
         |
         |class A {
         |  def m1: Unit = {
         |    def m: Unit = { // m$1
         |      println(1)
         |      def m(x: String): Unit = // m$2
         |        println(2)
         |      m("hello")
         |    }
         |    m
         |  }
         |
         |  def m2: Unit = {
         |    def m(x: Int): Unit = println(3) // m$3
         |    m(1)
         |  }
         |}
         |""".stripMargin
    val debuggee = TestingDebuggee.mainClass(source, "example", scalaVersion)
    val unpickler = getUnpickler(debuggee)
    unpickler.assertFormat("example.A", "void m$1()", "A.m1.m: Unit")
    unpickler.assertFormat("example.A", "void m$2(java.lang.String x)", "A.m1.m.m(x: String): Unit")
    unpickler.assertFormat("example.A", "void m$3(int x)", "A.m2.m(x: Int): Unit")
  }
```

The comments below shuld help you fix this.

modules/unpickler/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3Unpickler.scala
aymanelamyaghri: I added this for the case of a value class,   we find the method in the tree of the owner of the declaringType

adpi2: I see.

For the case of the value class you can search in the companion class of the `declaringType`. If you need help to find the symbol of the companion class, you can ask Jamie.

modules/unpickler/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3Unpickler.scala
```diff
@@ -57,17 +59,48 @@ class Scala3Unpickler(
...
+                    else
+                      if matchTargetName(name, symbol) then currentindex += 1
+                      List()
+
+                  case _ => List()
+              else List()
+            )((l1, l2) => l1 ++ l2, List())
+          case None => List()
+      )
+      .headOption
```
adpi2, line 103: The following would be slightly less performant but easier to read.

```suggestion
    val matchingSymbols = declaringtpe.declarations
      .flatMap(sym => sym.tree)
      .flatMap(tree =>
        tree.walkTree {
          case DefDef(_, _, _, _, symbol) if matchTargetName(name, symbol) => List(symbol)
          case ValDef(_, _, _, symbol) if matchTargetName(name, symbol) => List(symbol)
          case _ => List.empty
        }(_ ++ _, List.empty)
      )
    Some(matchingSymbols(index - 1))
```

modules/unpickler/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3Unpickler.scala
adpi2: With the current implement there will be a problem if two classes in the same package contain similar local methods:


```
package example

class A {
  def m: Unit = {
    def m: Unit = println("A")
    m
  }
}

class B {
    def m: Unit = {
      def m: Unit = println("A")
      m
  }
}
```

If we look for the local method `m` in `B`, we will find the one in `A`. It is really important to look in the correct `declaringType`.

So for the case of a value class, we can test if `declaringType` is the companion object of a value class (a class that extends `AnyVal`) and if it is we should return the companion class instead of the companion object. This should probably be done in `findDeclaringType` instead of this:
https://github.com/scalacenter/scala-debug-adapter/blob/084ce5436e697f9844813e58cd810c90433bfc21/modules/unpickler/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3Unpickler.scala#L213-L216C78

adpi2: For the local method in a value class, we receive its companion object as declaring type but we need to search in both the companion object and if not found in the companion class.

PR #514, opened 2023-07-17
added -g javac option & updated bin/scalafmt

### LINKED ISSUE #414: [sbt] Add -g to the javacOptions to obtain the debugging info

iusildra:
fixes #414


modules/sbt-plugin/src/main/scala/ch/epfl/scala/debugadapter/sbtplugin/DebugAdapterPlugin.scala
```diff
@@ -100,7 +100,12 @@ object DebugAdapterPlugin extends sbt.AutoPlugin {
   def runSettings: Seq[Def.Setting[_]] = Seq(
     startMainClassDebugSession := mainClassSessionTask.evaluated,
     startRemoteDebugSession := remoteSessionTask.evaluated,
-    stopDebugSession := stopSessionTask.value
+    stopDebugSession := stopSessionTask.value,
+    Keys.compile / Keys.javacOptions := {
+      val jo = (Keys.compile / Keys.javacOptions).value
+      if (jo.contains("-g")) jo
```
adpi2, line 106: According to the `javac` documentation:

> -g
Generates all debugging information, including local variables. By default, only line number and source file information is generated.
-g:[lines, vars, source]
Generates only the kinds of debugging information specified by the comma-separated list of keywords. Valid keywords are:
lines
Line number debugging information.
vars
Local variable debugging information.
source
Source file debugging information.
-g:none
Does not generate debugging information.

So if it contains any of `-g:foo, bar` we should probably not add `-g`.

```suggestion
      if (jo.contains("-g")) jo
```

modules/sbt-plugin/src/sbt-test/debug-session/debug-java/withG/src/main/java/example/Main.java
```diff
@@ -0,0 +1,8 @@
+package example;
+
+public class Main {
+  public static void main(String[] args) {
+    var x = 1;
+    System.out.println("Hello, world!");
+  }
+}
```
adpi2, line 8: You can add a new line at the end of this file. Same with the other `Main.java`

modules/sbt-plugin/src/sbt-test/debug-session/debug-java/withGSpecialized/src/main/java/example/Main.java
```diff
@@ -0,0 +1,12 @@
...
+
+public class Main {
+  public static void main(String[] args) {
+    int x = 1;
+    System.out.println("Hello, world!");
+  }
+
+  public static String foo() {
+    return "foo";
+  }
```
adpi2, line 11: This seems useless?

adpi2 (approved):
LGTM

Issue #486, opened 2023-07-05
Cannot evaluate expression in Scala 2.12 with sbt 1.9.0


adpi2:
Something is wrong with the classpath:
```
2023.07.05 17:03:44 ERROR java.lang.NoClassDefFoundError: scala/jdk/CollectionConverters$
	at scala.tools.nsc.ExpressionCompilerBridge.run(ExpressionCompilerBridge.scala:40)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at ch.epfl.scala.debugadapter.internal.evaluator.ExpressionCompiler.compile(ExpressionCompiler.scala:48)
	at ch.epfl.scala.debugadapter.internal.evaluator.ScalaEvaluator.$anonfun$compile$7(ScalaEvaluator.scala:47)
	at ch.epfl.scala.debugadapter.internal.evaluator.Safe.flatMap(Safe.scala:29)
	at ch.epfl.scala.debugadapter.internal.evaluator.ScalaEvaluator.$anonfun$compile$3(ScalaEvaluator.scala:44)
	at ch.epfl.scala.debugadapter.internal.evaluator.Safe.flatMap(Safe.scala:29)
	at ch.epfl.scala.debugadapter.internal.evaluator.ScalaEvaluator.compile(ScalaEvaluator.scala:43)
	at ch.epfl.scala.debugadapter.internal.EvaluationProvider.$anonfun$prepare$2(EvaluationProvider.scala:138)
	at scala.util.Success.flatMap(Try.scala:251)
	at ch.epfl.scala.debugadapter.internal.EvaluationProvider.$anonfun$prepare$1(EvaluationProvider.scala:135)
	at scala.util.Success.flatMap(Try.scala:251)
	at ch.epfl.scala.debugadapter.internal.EvaluationProvider.prepare(EvaluationProvider.scala:134)
	at ch.epfl.scala.debugadapter.internal.EvaluationProvider.evaluateForBreakpoint(EvaluationProvider.scala:77)
[...]

PR #515, opened 2023-07-17
Scala212 sbt190

iusildra:
fixes: #486


modules/sbt-plugin/src/main/scala/ch/epfl/scala/debugadapter/sbtplugin/DebugAdapterPlugin.scala
```diff
@@ -62,7 +62,9 @@ object DebugAdapterPlugin extends sbt.AutoPlugin {
     val startRemoteDebugSession = inputKey[URI](
       "Start a debug session on a remote process"
     ).withRank(KeyRanks.DTask)
-
+    val debugConfig = settingKey[DebugConfig](
```
adpi2, line 65: We should try to avoid name conflicts in all sbt plugins. `debugConfig` seems to be a pretty common name, so we should make it more likely to be unique. I suggest `debugAdapterConfig`.

```suggestion
    val debugAdapterConfig = settingKey[DebugConfig](
```

adpi2 (commented):
One small but important comment, otherwise it LGTM.

adpi2 (approved):

PR #516, opened 2023-07-18
find and  format local classes

aymanelamyaghri:


modules/unpickler/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3Unpickler.scala
```diff
@@ -237,24 +237,46 @@ class Scala3Unpickler(
...
-      .flatMap { sym =>
-        val encodedSymName = NameTransformer.encode(sym.name.toString)
-        val Symbol = s"${Regex.quote(encodedSymName)}\\$$?(.*)".r
-        encodedName match
-          case Symbol(remaining) =>
-            if remaining.isEmpty then Some(sym)
-            else findSymbolsRecursively(sym, remaining)
-          case _ => None
-      }
+    val pattern = """^([^$]+)[$](\d+)[$]?(.*)$""".r
```
adpi2, line 240: A few comments about the regex:
- I would rather use `\` to escape the `$`
- The final `$` is useless because `.*` matches everything up to the end
- The entire final `\$.*` is optional, not just the `\$` separator

```suggestion
    val pattern = """^([^$]+)\$(\d+)(\$.*)?$""".r
```

`remaining` would then start with `$` though. You can strip it with `remaining.stripPrefix("$")`

adpi2:
supersed by #527

PR #519, opened 2023-07-19
update-tasty-format-varargs-and-implicit/using

aymanelamyaghri:


modules/unpickler/src/test/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3UnpicklerTests.scala
```diff
@@ -208,6 +208,20 @@ abstract class Scala3UnpicklerTests(val scalaVersion: ScalaVersion) extends FunS
     unpickler.assertFormat("example.B", javaSig("java.lang.String"), "B.m(): String")
   }
 
+  test("using and implicit parameters") {
+    val source =
+      """|package example
+         |object Main{
+         |  def m1(using x : Int , y : Int) = x+y
```
adpi2, line 215: What if we don't give a name to the parameters:

```scala
         |  def m3(using String , Int): Unit = ()
```

adpi2 (approved):
One small test to add but apart from that LGTM

PR #521, opened 2023-07-20
Skip private accessors

### LINKED ISSUE #332: Should skip over package-private accessor in case class
For instance, in the standard library,
```scala
final case class :: [+A](override val head: A, private[scala] var next: List[A @uncheckedVariance])
```

The method `$colon$colon.next$access$1()` is the getter of field `next` and it should be skipped.

## How to fix

In `ScalaStepFilter` we can skip methods matching `$access$\d+`.

Can the field also be lazy? What should we do?

iusildra:
Fixes #332 
Only works for Scala 2 as there are no methods named `xxx$access$\d+` in Scala 3
Also remvoved `final` attribute from `FakeJdiMethod` and `FakeJdiType` to be able to create anon classes to solve `NotImplementedError` at call site

### DIFF
--- a/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/ScalaUnpickler.scala
+++ b/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/ScalaUnpickler.scala
@@ -24,12 +24,12 @@ abstract class ScalaUnpickler(scalaVersion: ScalaVersion, testMode: Boolean) ext
   override def shouldSkipOver(method: Method): Boolean = {
     if (method.isBridge) true
     else if (isDynamicClass(method.declaringType)) true
-    // else if (scalaVersion.isScala2 && isPackagePrivateAccessor(method)) true
     else if (isJava(method)) false
     else if (isConstructor(method)) false
     else if (isStaticConstructor(method)) false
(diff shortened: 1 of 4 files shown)


iusildra:
I should move the check `isPackagePrivate` check to `Scala2Unpickler`, wdyt ?

adpi2:
> I should move the isPackagePrivate check to Scala2Unpickler, wdyt ?

I think it is fine to keep it in `ScalaUnpickler` for now.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/ScalaUnpickler.scala
```diff
@@ -143,6 +144,9 @@ abstract class ScalaUnpickler(scalaVersion: ScalaVersion, testMode: Boolean) ext
 
   private def skipTraitInitializer(method: Method): Boolean =
     method.bytecodes.toSeq == Seq(ByteCodes.RETURN)
+
+  private def isPrivateAccessor(method: Method): Boolean =
+    method.name.matches(""".+\$access\$\d+""")
```
iusildra, line 149: I thought it did not work in Scala 2.12

iusildra (commented):
Looks good, but I can't approve 😆

PR #522, opened 2023-07-20
no transform when inline

### LINKED ISSUE #454: Add support for `inline` methods and fields
Scala 3 only

iusildra:
Fixes #454

### DIFF
--- a/modules/expression-compiler/src/main/scala-3/dotty/tools/dotc/evaluation/ExtractExpression.scala
+++ b/modules/expression-compiler/src/main/scala-3/dotty/tools/dotc/evaluation/ExtractExpression.scala
@@ -68,7 +68,12 @@ class ExtractExpression(using exprCtx: ExpressionContext) extends MacroTransform
       desugaredIdent match
         case tree: ImportOrExport => tree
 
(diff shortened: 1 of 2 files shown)


modules/expression-compiler/src/main/scala-3/dotty/tools/dotc/evaluation/ExtractExpression.scala
```diff
@@ -68,7 +68,12 @@ class ExtractExpression(using exprCtx: ExpressionContext) extends MacroTransform
       desugaredIdent match
         case tree: ImportOrExport => tree
 
-        case tree if tree.symbol.is(Inline) => tree
+        case tree if tree.symbol.is(Inline) =>
+          tree.symbol.info match
+            case ConstantType(value) => Literal(value)
```
adpi2, line 73: It is safer to use `cpy(tree).Literal(value)` to not lose all the attachements that we can have on `tree`.

```suggestion
            case ConstantType(value) => cpy(tree).Literal(value)
```

adpi2 (approved):
One small comment and LGTM.

Thanks for fixing this issue.

PR #524, opened 2023-07-21
allow step into in classes with special names

### LINKED ISSUE #508: Cannot pause inside inner class `&&` of class `A+B` in Scala 2
I have absolutely no idea of why in Scala 2 there are no breakpoints inside `A+B$$amp$amp$#x` in file `RuntimeEvaluatorTests.scala:898`
I launched the same program in a debug session and placed breakpoints on the whole method. Never reached them

_Originally posted by @iusildra in https://github.com/scalacenter/scala-debug-adapter/issues/484#issuecomment-1634414548_

iusildra:
Fixes #508

### DIFF
--- a/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala2Unpickler.scala
+++ b/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala2Unpickler.scala
@@ -1,12 +1,14 @@
 package ch.epfl.scala.debugadapter.internal.stacktrace
 
-import com.sun.jdi
-import ch.epfl.scala.debugadapter.internal.scalasig._
-import ch.epfl.scala.debugadapter.internal.SourceLookUpProvider
-import ch.epfl.scala.debugadapter.internal.scalasig.ScalaSigPrinter
 import ch.epfl.scala.debugadapter.Logger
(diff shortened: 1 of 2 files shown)


modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala2Unpickler.scala
```diff
@@ -124,7 +124,7 @@ class Scala2Unpickler(
       .foldRight(Option(javaClass.name)) { (sym, acc) =>
         for (javaName <- acc if javaName.contains(sym.name))
           yield javaName
-            .split(sym.name)
+            .split(sym.name.replaceAll("\\$", "\\\\\\$"))
```
iusildra, line 127: Damn, fells like programming on Windows

adpi2: I think you are trying to do `Regex.quote(sym.name)` ;)

```scala
            .split(Regex.quote(sym.name))
```

modules/java-debug
adpi2, line 1: @aymanelamyaghri and you are always pushing `java-debug` back and forth. But this one looks more up-to-date.

PR #525, opened 2023-07-21
Skip given lazy val getter

### LINKED ISSUE #333: Should skip some given instances
A given instance can be typed as:
- a lazy val: `given Foo[Int] = ???`
- a def: `given [A]: Foo[A] = ???` or `given [A]: Foo[A] with = ???`
- an object: `given Foo[Int] with { ... }`

If it is a lazy val we should skip the getter `given_Foo_Int()`

iusildra:
Fixes #333

### DIFF
--- a/modules/unpickler/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3Unpickler.scala
+++ b/modules/unpickler/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3Unpickler.scala
@@ -18,9 +18,6 @@ import java.nio.file.Path
 import java.util.Optional
 import java.util.function.Consumer
 import scala.jdk.OptionConverters.*
-import scala.util.Failure
-import scala.util.Success
-import scala.util.Try
 import scala.util.matching.Regex
 import tastyquery.Modifiers.TermSymbolKind
 import tastyquery.SourceLanguage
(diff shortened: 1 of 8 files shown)


modules/unpickler/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3Unpickler.scala
```diff
@@ -361,4 +361,6 @@ class Scala3Unpickler(
 
     val isNonLazyGetterOrSetter =
       (!symbol.isMethod || isScala2GetterWorkaround || symbol.isSetter) && symbol.kind != TermSymbolKind.LazyVal
-    isNonLazyGetterOrSetter || symbol.isSynthetic
+    val isGivenLazyVal =
+      symbol.isGivenOrUsing && symbol.kind == TermSymbolKind.LazyVal
```
adpi2, line 365: Why don't we skip all lazy getters and setters?

iusildra: Hum, could try and see what happens

adpi2:
Hey @iusildra, it seems this issue with lazy getters and initializers is harder to fix that I thought. I am working on it now, in this PR.

modules/unpickler/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3Unpickler.scala
adpi2: The reason why is that from Scala 3.0 to 3.2 there was a single method for getting and initializing the field. We now have 2 methods in Scala 3.3 and so we can skip the getter and not skip the initializer. As a downside it will always skip the getter-initializer in Scala 3.0 and 3.2 but I guess we can live with that. Scala 3.3 is the LTS after all.

aymanelamyaghri (approved):

PR #527, opened 2023-07-24
find-symbol-local-class

aymanelamyaghri:


modules/unpickler/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3Unpickler.scala
```diff
@@ -337,6 +388,8 @@ class Scala3Unpickler(
       scalaType match
         case "scala.Any[]" =>
           javaType == "java.lang.Object[]" || javaType == "java.lang.Object"
+        case "scala.PolyFunction" =>
+          javaType == "scala.Function1"
```
adpi2, line 392: A poly function can have more than one arguments and thus the erasure can be `Function1` or `Function2` etc. You should fix this and add a test.

modules/unpickler/src/test/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3UnpicklerTests.scala
```diff
@@ -229,6 +253,61 @@ abstract class Scala3UnpicklerTests(val scalaVersion: ScalaVersion) extends FunS
...
+         |    class G extends A :
+         |      def m() = ()
+         |  }
+         |  }
+         |""".stripMargin
+    val debuggee = TestingDebuggee.mainClass(source, "example.Main", scalaVersion)
+    debuggee.assertFormat("example.Main$C$1", "int m()", "Main.m.C.m(): Int")
+    debuggee.assertFormat("example.Main$E$1$F", "void m()", "Main.m.E.F.m(): Unit")
+    debuggee.assertFormat("example.Main$D$1", "void m1()", "Main.m.D.m1: Unit")
+    debuggee.assertFailure("example.Main$G$1", "void m()")
```
adpi2, line 290: We should probably simplify these tests by adding an `assertFormat` that only takes the class name and not the method signature. It would make it unneeded to add all those boring `def m()` methods.

modules/unpickler/src/test/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3UnpicklerTests.scala
```diff
@@ -435,6 +517,7 @@ abstract class Scala3UnpicklerTests(val scalaVersion: ScalaVersion) extends FunS
          |  val x: A = new A {}
          |  def m(a: x.type)(b: x.B): A = a
          |  def m(t: Int !: Int) = 1
+         |  def m4() : [T] => Foo => Bar = ???
```
adpi2, line 520: This one looks more realistic:

```suggestion
         |  def m() : [T] => List[T] => Option[T] = ???
```

modules/unpickler/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/Scala3Unpickler.scala
```diff
@@ -246,39 +259,75 @@ class Scala3Unpickler(
...
+    assert(objSym.size <= 1 && clsSym.size <= 1)
+    if cls.isObject && !isExtensionMethod then objSym.headOption else clsSym.headOption
+
+  object LocalClass:
+    def unapply(name: String): Option[(String, String, Option[String])] =
+      "(.+)\\$([^$]+)\\$\\d+(\\$.*)?".r
+        .unapplySeq(name)
+        .map(xs => (xs(0), xs(1), Option(xs(2)).map(_.stripPrefix("$")).filter(_.nonEmpty)))
 
   private def findSymbolsRecursively(owner: DeclaringSymbol, encodedName: String): Seq[ClassSymbol] =
```
aymanelamyaghri, line 288:
```suggestion
  private def findSymbolsRecursively(owner: DeclaringSymbol, decodedName: String): Seq[ClassSymbol] =
```

PR #542, opened 2023-07-31
Runtime assign

### LINKED ISSUE #443: Assign value of local var or field in runtime evaluator

iusildra:
Fixes #443

### DIFF
--- a/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeDefaultEvaluator.scala
+++ b/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeDefaultEvaluator.scala
@@ -22,6 +22,7 @@ class RuntimeDefaultEvaluator(val frame: JdiFrame, implicit val logger: Logger) 
       case array: ArrayElemTree => evaluateArrayElement(array)
       case branching: IfTree => evaluateIf(branching)
       case staticMethod: StaticMethodTree => invokeStatic(staticMethod)
+      case assign: AssignTree => evaluateAssign(assign)
       case UnitTree => Safe(JdiValue(frame.thread.virtualMachine.mirrorOfVoid, frame.thread))
     }
 
@@ -118,6 +119,26 @@ class RuntimeDefaultEvaluator(val frame: JdiFrame, implicit val logger: Logger) 
       predicate <- eval(tree.p).flatMap(_.unboxIfPrimitive).flatMap(_.toBoolean)
       value <- if (predicate) eval(tree.thenp) else eval(tree.elsep)
     } yield value
+
+  /* -------------------------------------------------------------------------- */
+  /*                              Assign evaluation                             */
+  /* -------------------------------------------------------------------------- */
+  def evaluateAssign(tree: AssignTree): Safe[JdiValue] = {
+    eval(tree.rhs)
+      .flatMap { rhsValue =>
+        tree.lhs match {
+          case field: InstanceFieldTree =>
+            eval(field.qual).flatMap { qualValue =>
+              Safe(qualValue.asObject.reference.setValue(field.field, rhsValue.value))
+            }
+          case field: StaticFieldTree => Safe(field.on.setValue(field.field, rhsValue.value))
+          case localVar: LocalVarTree =>
+            val localVarRef = frame.variableByName(localVar.name)
+            Safe(localVarRef.map(frame.setVariable(_, rhsValue)).get)
+        }
+      }
+      .map(_ => JdiValue(frame.thread.virtualMachine.mirrorOfVoid, frame.thread))
+  }
 }
 
 object RuntimeDefaultEvaluator {
--- a/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeDefaultEvaluator.scala
+++ b/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeDefaultEvaluator.scala
@@ -128,8 +128,8 @@ class RuntimeDefaultEvaluator(val frame: JdiFrame, implicit val logger: Logger) 
       .flatMap { rhsValue =>
         tree.lhs match {
           case field: InstanceFieldTree =>
-            eval(field.qual).flatMap { qualValue =>
-              Safe(qualValue.asObject.reference.setValue(field.field, rhsValue.value))
+            eval(field.qual).map { qualValue =>
+              qualValue.asObject.reference.setValue(field.field, rhsValue.value)
             }
           case field: StaticFieldTree => Safe(field.on.setValue(field.field, rhsValue.value))
           case localVar: LocalVarTree =>
--- a/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeDefaultValidator.scala
+++ b/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeDefaultValidator.scala
@@ -43,6 +43,7 @@ class RuntimeDefaultValidator(val frame: JdiFrame, val sourceLookUp: SourceLookU
       case branch: Term.If => validateIf(branch)
       case instance: Term.New => validateNew(instance)
       case block: Term.Block => validateBlock(block)
+      case assign: Term.Assign => validateAssign(assign)
       case _ => Recoverable("Expression not supported at runtime")
     }
 
@@ -205,13 +206,6 @@ class RuntimeDefaultValidator(val frame: JdiFrame, val sourceLookUp: SourceLookU
       result <- validateApply(intermediate, args)
     } yield result
 
-  /*
-   * validateIndirectApply MUST be called before methodTreeByNameAndArgs
-   * That's because at runtime, a inner module is accessible as a 0-arg method,
-   * and if its associated class has not attribute either, when calling the
-   * constructor of the class, methodTreeByNameAndArgs would return its companion
-   * object instead. Look at the test about multiple layers for an example
-   */
   def findMethod(
       tree: RuntimeTree,
       name: String,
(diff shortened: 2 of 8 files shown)


modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeDefaultEvaluator.scala
```diff
@@ -118,6 +119,26 @@ class RuntimeDefaultEvaluator(val frame: JdiFrame, implicit val logger: Logger)
...
+  /* -------------------------------------------------------------------------- */
+  /*                              Assign evaluation                             */
+  /* -------------------------------------------------------------------------- */
+  def evaluateAssign(tree: AssignTree): Safe[JdiValue] = {
+    eval(tree.rhs)
+      .flatMap { rhsValue =>
+        tree.lhs match {
+          case field: InstanceFieldTree =>
+            eval(field.qual).flatMap { qualValue =>
+              Safe(qualValue.asObject.reference.setValue(field.field, rhsValue.value))
```
adpi2, line 132:
```suggestion
            eval(field.qual).map { qualValue =>
              qualValue.asObject.reference.setValue(field.field, rhsValue.value)
```

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeDefaultValidator.scala
```diff
@@ -304,6 +297,39 @@ class RuntimeDefaultValidator(val frame: JdiFrame, val sourceLookUp: SourceLookU
...
+      localVarValidation: String => Validation[RuntimeEvaluableTree] = localVarTreeByName,
+      fieldValidation: (Validation[RuntimeTree], String) => Validation[RuntimeEvaluableTree] = fieldTreeByName
+  ): Validation[RuntimeEvaluableTree] = {
+    val lhs = tree.lhs match {
+      case select: Term.Select =>
+        fieldValidation(validateWithClass(select.qual), select.name.value)
+      case name: Term.Name =>
+        localVarValidation(name.value).orElse(fieldValidation(thisTree, name.value))
+      case _ => Recoverable("Unsupported assignment")
+    }
```
adpi2, line 322: I wonder if it would not be simpler to pass a boolean to `filedTreeByName` and `localVarTreeByName` to prevent them from pre-evaluating:

```suggestion
  def validateAssign(tree: Term.Assign): Validation[RuntimeEvaluableTree] = {
    val lhs = tree.lhs match {
      case select: Term.Select =>
        fieldTreeByName(validateWithClass(select.qual), select.name.value, preEvaluate = false)
      case name: Term.Name =>
        localVarTreeByName(name.value, preEvaluate = false).orElse(fieldTreeByName(thisTree, name.value, preEvaluate = false))
      case _ => Recoverable("Unsupported assignment")
    }
```

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/internal/RuntimeEvaluatorTests.scala
```diff
@@ -963,6 +963,56 @@ abstract class RuntimeEvaluatorTests(val scalaVersion: ScalaVersion) extends Deb
...
+         |  def main(args: Array[String]): Unit = {
+         |    val a = new A
+         |    var b = 0
+         |    var bc: B = new C
+         |    println("ok")
+         |  }
+         |}
+         |""".stripMargin
+    implicit val debuggee: TestingDebuggee = TestingDebuggee.mainClass(source, "example.Main", scalaVersion)
+    def assignSuccess(expr: String) = Evaluation.success(expr)(res => res == "<void value>")
```
adpi2, line 994: Would it be possible to write it this way?

```suggestion
    def assignSuccess(expr: String) = Evaluation.success(expr, ())
```

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/internal/RuntimeEvaluatorTests.scala
```diff
@@ -963,6 +963,56 @@ abstract class RuntimeEvaluatorTests(val scalaVersion: ScalaVersion) extends Deb
...
+    check(
+      Breakpoint(21),
+      assignSuccess("a.a = 42"),
+      Evaluation.success("a.a", 42),
+      assignSuccess("a.f.a = 64"),
+      Evaluation.success("a.f.a", 64),
+      assignSuccess("b = 42"),
+      Evaluation.success("Main.b", 0),
+      Evaluation.successOrIgnore("b", 42, true), // ? works in a debug session but not in tests ???
```
adpi2, line 1003: :thinking: ?

```suggestion
      Evaluation.successOrIgnore("b", 42, true), // ? works in a debug session but not in tests ???
```

iusildra: Yeah, that one's pretty weird. I doesn't work because their is a field `b` and a local variable `b`. If I remove the field, it works as expected.
But when I do a `publishLocal` and open a debug session, it works correctly :thinking:

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/internal/RuntimeEvaluatorTests.scala
iusildra: Nop, because I return a `mirrorOfVoid` after the assignment

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeDefaultValidator.scala
iusildra: Yep, seems simpler

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/internal/RuntimeEvaluatorTests.scala
adpi2: Could it be that in a debug session it falls back to the expression compiler?

adpi2: In the test, the precedence of the local variable should be greater than the field. It is probably worth fixing this.

iusildra: No, I blocked the Scala evaluation. But in fact it does not work in a debug session either but that's weird. In the local variable panel, I see the updated value, but when evaluating it I still got the old one (the code I used is the same as in the tests)
image

adpi2: Do we cache the `LocalVariable` anywhere?

iusildra: it was just because of a wrong evaluation order :smiling_face_with_tear:

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/internal/RuntimeEvaluatorTests.scala
```diff
@@ -1013,6 +1013,26 @@ abstract class RuntimeEvaluatorTests(val scalaVersion: ScalaVersion) extends Deb
       Evaluation.failed("Main.b = \"str\"")
     )
   }
+
+  test("local var should not mask field of same name (and vice-versa)") {
```
adpi2, line 1017:
```suggestion
  test("a field should not shadow a local variable") {
```

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/internal/RuntimeEvaluatorTests.scala
```diff
@@ -1013,6 +1013,26 @@ abstract class RuntimeEvaluatorTests(val scalaVersion: ScalaVersion) extends Deb
...
+         |    println(a)
+         |  }
+         |}
+         |""".stripMargin
+    implicit val debuggee: TestingDebuggee = TestingDebuggee.mainClass(source, "example.Main", scalaVersion)
+    check(
+      Breakpoint(7),
+      Evaluation.success("a", 42),
+      Evaluation.success("Main.a", 0)
+    )
```
adpi2, line 1034: You should probably add:

```scala
Evaluation.success("this.a", 0)
```

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeDefaultValidator.scala
```diff
@@ -196,29 +194,24 @@ class RuntimeDefaultValidator(val frame: JdiFrame, val sourceLookUp: SourceLookU
...
   def validateIndirectApply(
-      on: Validation[RuntimeTree],
+      on: RuntimeTree,
       name: String,
       args: Seq[RuntimeEvaluableTree]
   ): Validation[RuntimeEvaluableTree] =
     for {
-      intermediate <- validateName(name, on, methodFirst = true).orElse(validateClass(name, on))
+      intermediate <- validateMember(name, on, methodFirst = true)
+        .orElse(localVarTreeByName(name))
```
adpi2, line 203: Do we still need this?

```suggestion
        .orElse(localVarTreeByName(name))
```

iusildra: Yes, because the "indirect" apply can be either on a field/method or on a local var (e.g. `list(0)`)

adpi2: Ok, I see

adpi2 (approved):

PR #544, opened 2023-08-01
Re-use hot filesystem to read sources

### LINKED ISSUE #541: Debugging very slow to start
When trying to debug [this test](https://github.com/scalacenter/gradle-bloop/blob/1490802fcfbadaf856cd4957e97d490041b2e191/src/test/scala/bloop/integrations/gradle/ConfigGenerationSuite.scala#L31) using Bloop/Metals/VSCode/Windows, I find the debugger very slow to start (> 5 mins)

This is mainly because of https://github.com/scalacenter/scala-debug-adapter/blob/faccb7f7810903ae1c37ee97f87e897d84340b50/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/ClassEntryLookUp.scala#L236

It's being hit many times on the gradle 5.0 jar: https://repo1.maven.org/maven2/dev/gradleplugins/gradle-api/5.0/

The Gradle jar and its sources are pretty busy (>3K folders, >12K files) and source files often don't exist for the corresponding class file.

The only class that I've been able to pin down as constantly being searched for is `package-info.class`/`package-info.java`.  Can this just be excluded from this area of code?  AFAIK - there's never any runnable code in these files.

Another possible performance improvement could be to group these missing files by `jar` and then scan for all of them instead of opening/closing the jar FS for every class file.

adpi2:
Fix #541 

On my laptop it speeds up the loading of [gradle-api](https://repo1.maven.org/maven2/dev/gradleplugins/gradle-api/5.0/) from 140s to 1.8s.

### DIFF
--- a/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/ClassEntryLookUp.scala
+++ b/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/ClassEntryLookUp.scala
@@ -160,13 +160,13 @@ private class ClassEntryLookUp(
 
 private object ClassEntryLookUp {
   private[internal] def apply(entry: ClassEntry, logger: Logger): ClassEntryLookUp = {
-    val sourceFiles = entry.sourceEntries.flatMap(SourceEntryLookUp.getAllSourceFiles(_, logger))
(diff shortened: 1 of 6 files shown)


Arthurm1:
That's great!  Thanks @adpi2

PR #570, opened 2023-08-17
Sync HCR with compilation process

iusildra:


adpi2 (approved):
Is this ready to merge?

iusildra:
It should be, just need to check a few things beforehand

PR #594, opened 2023-09-15
Skip exported methods and BoxeRuntimes

### LINKED ISSUE #560: Skip the exported method in Scala 3
```scala
class Foo {
 def foo = println(foo)
}

class Bar(foo: Foo) {
  export foo.*
}

@main def test = {
  val foo = new Foo
  val bar = new Bar(foo)
  bar.foo // <= step into should go directly to Foo.foo
}
```

iusildra:
Fixes #560 #540

### DIFF
--- a/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/RuntimeStepFilter.scala
+++ b/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/RuntimeStepFilter.scala
@@ -5,13 +5,10 @@ import com.sun.jdi.Location
 import ch.epfl.scala.debugadapter.ScalaVersion
 
 private[internal] class RuntimeStepFilter(classesToSkip: Set[String], methodsToSkip: Set[String]) extends StepFilter {
-  override def shouldSkipOver(method: Method): Boolean = shouldSkipOut(method) || isBoxingMethod(method)
+  override def shouldSkipOver(method: Method): Boolean = shouldSkipOut(method)
   override def shouldSkipOut(upperLocation: Location, method: Method): Boolean = shouldSkipOut(method)
   private def shouldSkipOut(method: Method): Boolean =
     classesToSkip.contains(method.declaringType.name) || methodsToSkip.contains(method.toString)
-  private def isBoxingMethod(method: Method): Boolean = {
-    method.declaringType().name().startsWith("scala.runtime.BoxesRunTime")
-  }
 }
 
 private[internal] object RuntimeStepFilter {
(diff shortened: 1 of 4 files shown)


modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/ScalaUnpickler.scala
```diff
@@ -24,6 +24,7 @@ abstract class ScalaUnpickler(scalaVersion: ScalaVersion, testMode: Boolean) ext
 
   override def shouldSkipOver(method: Method): Boolean = {
     if (method.isBridge) true
+    else if (isBoxingMethod(method)) true
```
adpi2, line 27: We have a different class for filtering out the runtime methods which is the `RuntimeStepFilter`. This should be implemented there instead of in the `ScalaUnpickler`.

adpi2 (commented):
What about the exported methods. Are they already filtered out? What about adding a test? Should we work on them in a different PR?

iusildra:
They are already filtered out in the first commit (with a test), did I miss some exported methods?

adpi2:
> They are already filtered out in the first commit (with a test), did I miss some exported methods?

My bad. I looked at the last commit only.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/stacktrace/RuntimeStepFilter.scala
```diff
@@ -5,10 +5,13 @@ import com.sun.jdi.Location
 import ch.epfl.scala.debugadapter.ScalaVersion
 
 private[internal] class RuntimeStepFilter(classesToSkip: Set[String], methodsToSkip: Set[String]) extends StepFilter {
-  override def shouldSkipOver(method: Method): Boolean = shouldSkipOut(method)
+  override def shouldSkipOver(method: Method): Boolean = shouldSkipOut(method) || isBoxingMethod(method)
```
adpi2, line 8: I think you only need to add `"scala.runtime.BoxesRunTime"` to the set of `classesToSkip`.

adpi2:
There is a test that fails because it was stepping into `BoxesRunTime`:
```
[error] ==> X ch.epfl.scala.debugadapter.Scala212StepFilterTests.should match all kinds of types  3.13s munit.ComparisonFailException: /home/runner/work/scala-debug-adapter/scala-debug-adapter/modules/tests/src/test/scala/ch/epfl/scala/debugadapter/StepFilterTests.scala:775
[error] 774:      Breakpoint(32),
[error] 775:      StepIn.method("BoxesRunTime.boxToInteger(int): Integer"),
[error] 776:      StepOut.line(32),
[error] values are not the same
[error] => Obtained
[error] Main$.m(Object): Object
[error] => Diff (- obtained, + expected)
[error] -Main$.m(Object): Object
[error] +BoxesRunTime.boxToInteger(int): Integer
```

adpi2 (approved):
Thanks!

PR #609, opened 2023-10-25
Add step filter configuration

### LINKED ISSUE #561: Add configuration to enable/disable step filters
We currently have 3 step filters:
- the runtime step filter
- the classloading step filter
- the Scala step filter

The idea is to add configuration in Metals settings to enable or disable them.

After that, we can add more step filters that are disabled by default but could be enabled by the user: for instance the implicit conversions, the using parameters...

iusildra:
Fixes #561
At first I thought of passing an enum when starting the debug server, but still have to find out how to pass it from metals.
The idea is to have a set of class of step filter, and add them in the stack trace provider

The best would be to be able to change it in a debug session but I did not find any commands in java-debug to do so


adpi2:
> The best would be to be able to change it in a debug session but I did not find any commands in java-debug to do so

We can pass arguments through the `launch` request:
https://github.com/scalacenter/scala-debug-adapter/blob/08dbf4bcea86cdef5ef50e418ad6310dc2e4016f/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/DebugSession.scala#L154-L157

That's how `java-debug` does it so it's probably a good idea.

iusildra:
It seems to be working (at least a bit :upside_down_face:). Still has to create tests for it, but when toggling configs in vs-code, the received filters list is changing
I'll investigate a crash when:
1. start a debug session
2. toggle a filter
3. restart the session (restart, not stop then start)

iusildra:
Not ready to be merged yet, I just need some clearance. Among the 3 current step filters, only the runtime one seems relevant to me, because I can't think of a case where:
- one will be interested in the class loading
- a user will want to go through every bridges, accessors etc...
WDYT ?

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/PartialLaunchArguments.scala
```diff
@@ -1,5 +1,6 @@
 package ch.epfl.scala.debugadapter.internal;
 
 class PartialLaunchArguments {
-  public boolean noDebug = false;
-}
\ No newline at end of file
+  val noDebug = false
+  val scalaStepFilters = allFilters
```
adpi2, line 5: To answer your question, I think we should give the maximum flexibility to the user. Some users may want to see the raw behavior of their code at the JVM level, to do micro-optimization, or to debug the compiler...

However, I wonder if an array is the best shape for this. Maybe we should have a list of individual booleans:
```scala
val skipClassLoading: Boolean = true
val skipScalaRuntime: Boolean = true
// bridges, forwarders and accessors
val skipScalaGeneratedMethods: Boolean = true
```

Also now that I merged the better stack traces PR we could have:
```scala
val decodeStackTraces: Boolean = true
```

adpi2: Also we can have:

```scala
val enableScalaEvaluation: Boolean = true
val enableRuntimeEvaluation: Boolean = true
```

That would be super convenient for us to debug the evaluation.

adpi2:
Hey @iusildra I am looking forward to this PR! Do you know if you will have time to work on it? Or do you prefer me to take it over?

iusildra:
School's ending this Wednesday, I'll have some time to finish it 🙂
(I miss being able to work full-time on it 🥲)

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/PartialLaunchArguments.scala
```diff
@@ -0,0 +1,22 @@
...
+  def extractStepFilers(obj: JsonObject): Map[String, Boolean] = {
+    val customStepFilters = new ju.HashMap[String, Boolean]()
+    val stepFilters = obj.getAsJsonObject("scalaStepFilters")
+    for (entry <- stepFilters.entrySet().toArray) {
+      val key = entry.asInstanceOf[ju.Map.Entry[String, JsonObject]].getKey
+      val value = stepFilters.get(key).getAsBoolean
+      customStepFilters.put(key, value)
+    }
+    customStepFilters.asScala.toMap
+  }
```
iusildra, line 21: I need the further check the docs of `gson` because this is awful. Without this I get an empty map...

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/StepFilterTests.scala
```diff
@@ -2,14 +2,326 @@ package ch.epfl.scala.debugadapter
...
+         |  }
+         |}
+         |""".stripMargin
+    implicit val debuggee: TestingDebuggee = TestingDebuggee.mainClass(source, "example.Main", scalaVersion)
+    check(Array(internal.classLoadingFilterName, internal.runtimeFilterName))(
+      Breakpoint(12),
+      StepIn.line(8),
+      StepIn.line(3),
+      StepIn.line(5)
+    )
```
iusildra, line 29: Contrary to Scala 2, there is a debug point at the trait's definition. But there's a weird method:
```sh
Stepping in, at Main$.main(String[]): void
Stepping in, at Fooo.m(): int
Stepping in, at Foo.m$(Foo): int // <-- what is this method ?
```

iusildra: Ok tha'ts a static forwarder :grin:

modules/tests/src/test/scala/ch/epfl/scala/debugadapter/StepFilterTests.scala
```diff
@@ -2,14 +2,155 @@ package ch.epfl.scala.debugadapter
...
+         |""".stripMargin
+    implicit val debuggee: TestingDebuggee = TestingDebuggee.mainClass(source, "example.Main", scalaVersion)
+    check(
+      DebugConfig.default.copy(
+        stepFilters = UsedStepFilters(decoder = true, runtime = false, classLoading = true)
+      )
+    )(
+      Breakpoint(9),
+      StepIn.method("ScalaRunTime.wrapRefArray[T](xs: Array[T]): ArraySeq[T]")
+    )
```
adpi2, line 30: It looks like this test is copied 3 times. To avoid that you can do:
```scala
val expectedFrame =
  if (scalaVersion.isScala3) "ScalaRunTime.wrapRefArray[T](xs: Array[T]): ArraySeq[T]"
  else if (scalaVersion.isScala213) "ScalaRunTime$.wrapRefArray(Object[]): ArraySeq"
  else "LowPriorityImplicits.wrapRefArray(Object[]): WrappedArray"
```

modules/tests/src/main/scala/ch/epfl/scala/debugadapter/testfmk/DebugTestSuite.scala
```diff
@@ -75,7 +78,9 @@ trait DebugTest extends CommonUtils {
     val client = TestingDebugClient.connect(server.uri)
     try {
       server.connect()
-      runAndCheck(client, attach = None, closeSession = true)(steps*)
+      runAndCheck(client, attach = None, closeSession = true, stepFilters = config.stepFilters)(
+        steps*
+      ) // ! That's a bit ugly because we need to pass it to LaunchArguments as well as DebugConfig but LaunchArguments' filters 'overrides' DebugConfig's filters.
```
adpi2, line 83: I don't think we need to pass the `stepFilters` to the `runAndCheck` method here. It should already be configured by the `config` via the "build tool".

```suggestion
      runAndCheck(client, attach = None, closeSession = true, stepFilters = config.stepFilters)(
        steps*
      ) // ! That's a bit ugly because we need to pass it to LaunchArguments as well as DebugConfig but LaunchArguments' filters 'overrides' DebugConfig's filters.
```

modules/tests/src/main/scala/ch/epfl/scala/debugadapter/testfmk/TestingDebugClient.scala
```diff
@@ -47,8 +49,12 @@ class TestingDebugClient(socket: Socket, logger: Logger)(implicit
     Await.result(sendRequest(request), timeout)
   }
 
-  def launch(timeout: Duration = defaultTimeout(16.seconds)): Messages.Response = {
-    val request = createRequest(Command.LAUNCH, new LaunchArguments())
+  def launch(
+      timeout: Duration = defaultTimeout(16.seconds),
+      stepFilters: UsedStepFilters = UsedStepFilters.default
```
adpi2, line 54:
```suggestion
      stepFilters: UsedStepFilters = null
      timeout: Duration = defaultTimeout(16.seconds),
```

modules/tests/src/main/scala/ch/epfl/scala/debugadapter/testfmk/DebugTestSuite.scala
```diff
@@ -94,13 +99,18 @@ trait DebugTest extends CommonUtils {
     }
   }
 
-  private def runAndCheck(client: TestingDebugClient, attach: Option[Int], closeSession: Boolean)(
+  private def runAndCheck(
+      client: TestingDebugClient,
+      attach: Option[Int],
+      closeSession: Boolean,
+      stepFilters: UsedStepFilters = UsedStepFilters.default
```
adpi2, line 106:
```suggestion
      stepFilters: UsedStepFilters = null
```

adpi2 (approved):
LGTM

PR #610, opened 2023-10-25
added a skip-ci

iusildra:
I thought it might be useful to be able to skip the tests (when only doc is changed or when we're on a "WIP WIP branch"), to avoid running them unnecessarily

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,15 +3,28 @@ on:
   push:
     branches: [main]
   pull_request:
+    paths:
+      - "modules/**"
+      - "project/**"
+      - ".scalafmt.conf"
+      - ".github/**"
+      - "build.sbt"
+      - "git-hook.sbt"
+
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 jobs:
   scalafmt:
+    if: ${{!contains(github.event.head_commit.message, '[skip ci]') || (github.event_name == 'pull_request' && !contains(github.event.pull_request.body, '[skip ci]'))}}
     name: Check Formatting
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - run: ./bin/scalafmt --test
   cross-compilation:
+    if: ${{!contains(github.event.head_commit.message, '[skip ci]') || (github.event_name == 'pull_request' && !contains(github.event.pull_request.body, '[skip ci]'))}}
     name: Cross Compilation
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -23,6 +23,7 @@ jobs:
           apps: sbt
       - run: sbt +Test/compile
   test:
+    if: ${{!contains(github.event.head_commit.message, '[no ci]')}}
```
adpi2, line 26: Could we get that from the name of the PR instead?

adpi2 (commented):
Thanks, that's a good idea.

Additionally we can trigger the CI only if some files changed. See for instance: https://github.com/scalacenter/course-management-tools/blob/abf583983225430a5d78b33ad003e5f7903676d7/.github/workflows/publish-docs.yml#L5-L8

Don't you think we should also disable the formatting and compilation jobs?

.github/workflows/ci.yml
iusildra: good point, but having it in commits can also be useful, don't you think ? (when we're pushing to save or other cases)
We could use both

adpi2: Yes it makes sense

iusildra:
Argl, I can't cancel workflows anymore :laughing:

tgodzik:
> Argl, I can't cancel workflows anymore 😆 As for formatting and compilation, yeah we could skip them too

I cancelled the previous ones, might be good to do it automatically like in Metals 

https://github.com/scalameta/metals/blob/main/.github/workflows/ci.yml#L8

tgodzik:
Or better yet https://github.com/VirtusLab/scala-cli/blob/main/.github/workflows/ci.yml#L11C1-L12C27

iusildra:
I also took some of dotty's ci config :D

.github/workflows/ci.yml
```diff
@@ -3,15 +3,28 @@ on:
...
+      - "build.sbt"
+      - "git-hook.sbt"
+
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 jobs:
   scalafmt:
+    if: ${{!contains(github.event.head_commit.message, '[skip ci]') || (github.event_name == 'pull_request' && !contains(github.event.pull_request.body, '[skip ci]'))}}
```
tgodzik, line 20: skip ci should work correctly by default in github https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/

Not sure why it didn't work for you :thinking:

iusildra: yep, I saw it too, but I think it will work only if [skip ci] is in the last commit title and not in the PR's

iusildra:
> Not sure why it didn't work for you 🤔

Seems to be it, it's build-in for commits message, but not for PR's title

tgodzik:
> > Not sure why it didn't work for you 🤔
> 
> Seems to be it, it's build-in for commits message, but not for PR's title

Ach yes, I guess it's that.

adpi2 (approved):
LGTM

Issue #614, opened 2023-11-14
SetBreakpoints command is not accommodating case-insensitive file paths appropriately


crt-31:
On Windows (and probably Mac), a DAP SetBreakpoints command with URI where the file-path’s casing is different than what is originally registered causes the breakpoint to be returned as ‘unverified’.

For instance: If I have a file C:\MyProject\A.scala, and send a SetBreakpoints command with source.path = “file:///C:/myproject/A.scala”,  it will return as unverified.  On Windows, this command should succeed. 

This issue is likely the cause of scalameta/Metals issue [#3747](https://github.com/scalameta/metals/issues/3747)

PR #615, opened 2023-11-14
Make ClassEntryLookup accommodate case-insensitive lookups where appropriate

crt-31:
This PR changes ClassEntryLookup so that it can lookup sourceFile URIs in a case-insensitive on case-insensitive file-systems.

For instance, on Windows, "file:///C/MyProject/A.scala" should map to the same objects as "file:///C:/myProject/A.scala". 

This PR should fix:
* This should fix scalameta/Metals issue [#3747](https://github.com/scalameta/metals/issues/3747)
* #614 

Solution Notes: 
- The maps in ClassEntryLookup now are keyed with a new SourceFileKey class that abstracts the URI and how it is represented.
- For Windows/Mac: 
  - For “file” scheme URIs, the comparison of the file part of the URI is compared in a case-insensitive way.
  - For “jar” scheme URIs, the comparison of the file part of the URI is compared in a case-insensitive way, but the jar-contents part is compared in case-sensitive way per jar rules. 
  - Other schemes are left case-sensitive
- Linux:
  - Remains case-sensitive

Since this is causing issues in Metals and Bloop, this PR should probably get into a release branch when approved.

### DIFF
--- a/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/ClassEntryLookUp.scala
+++ b/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/ClassEntryLookUp.scala
@@ -1,21 +1,20 @@
 package ch.epfl.scala.debugadapter.internal
 
 import ch.epfl.scala.debugadapter._
+import ch.epfl.scala.debugadapter.internal.ScalaExtension.*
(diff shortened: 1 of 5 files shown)


adpi2 (approved):
Thanks for the great fix. I'll release asap after merging.

PR #616, opened 2023-11-16
Log debuggee command with info level

adpi2:
Fixes https://github.com/scalameta/metals-feature-requests/issues/352 when sbt is the build server.


tgodzik (approved):
Ach ok, that's indented in the other PR as well. LGTM

PR #679, opened 2024-02-29
Allow lookup of class members in parents

### LINKED ISSUE #649: [Runtime Evaluation] Create instance of a member class of a parent class
```scala
class A:
  class B

class C extends A

val c: C = ???
```

It should be possible to evaluate `new c.B`. It fails with `Cannot find class B in qualifier C`.

iusildra:
Fixes #649


modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeValidation.scala
```diff
@@ -189,21 +190,58 @@ private[evaluator] class RuntimeValidation(frame: JdiFrame, sourceLookUp: Source
...
   private def findStaticClass(name: String, qualifier: jdi.ReferenceType): Validation[RuntimeClass] =
-    findQualifiedClass(name, qualifier.name)
+    findQualifiedClass(name)(qualifier.name)
       .filter(_.isStatic, cls => s"${cls.name} is not a static class")
       .map(StaticOrTopLevelClass.apply)
 
-  private def findMemberClass(name: String, qualifier: RuntimeEvaluationTree): Validation[RuntimeClass] =
-    findQualifiedClass(name, qualifier.`type`.name)
+  private def findClassInParent(name: String, startingPoint: RuntimeEvaluationTree): Validation[RuntimeClass] = {
+    var typesTraversed = List[jdi.ReferenceType]()
```
iusildra, line 198: I may should have used a `Queue` here :thinking:

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeValidation.scala
```diff
@@ -189,21 +190,58 @@ private[evaluator] class RuntimeValidation(frame: JdiFrame, sourceLookUp: Source
...
+        case cls: jdi.InterfaceType => cls.superinterfaces().asScala
+        case _ => Buffer()
+      }
+      // Reverse iterator to respect linearization
+      parentTypes.reverseIterator.withFilter(!typesTraversed.contains((_))).find { cls =>
+        typesTraversed = cls +: typesTraversed
+        ok = possibleClasses(cls.name)
+        ok.orElse(loop(cls)).isValid
+      }
+      ok
```
iusildra, line 219: That one's pretty ugly, but I'd like not to re-traverse a 'visited' type for performance

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/evaluator/RuntimeValidation.scala
adpi2: I rewrote this to functional style:
- First get all distinct parents as a Seq
- Then find the first matching inner class of a parent

In terms of performance, it is probably good enough.

PR #704, opened 2024-05-06
improvement: Use existing file system when available and don't close it

kasiaMarek:
connected to: https://github.com/scalameta/metals/issues/5928
We want to start scala debug adapter from Metals for build servers that don't support it. In Metals FileSystems for jars are created and kept open to be reused.


adpi2:
The [failing tests on macOS-latest](https://github.com/scalacenter/scala-debug-adapter/actions/runs/8968566990/job/24628685410?pr=704) are flaky tests, you don't need to fix them.

adpi2 (approved):
Thanks

PR #733, opened 2024-07-24
Fix CI job on main for 3.5+

rochala:


tgodzik (approved):

tgodzik:
The failure looks flaky, so I will go ahead and merge

.github/workflows/release-expression-compiler.yml
```diff
@@ -27,28 +27,32 @@ jobs:
   publish:
     if: startsWith(github.ref, 'refs/tags/v')
     runs-on: ubuntu-latest
+    needs: [test]
```
adpi2, line 30: I think we should remove this. In a new Scala 3 version it is likely that some evaluation tests fail but we still want to release the expression compiler and fix the evaluation on main.

rochala: Compilation will be tested anyway in publish step, so any binary incompatibilities will be found. If you think that this is enough, let's remove this

PR #742, opened 2024-08-06
Reverse entries in `SourceLookUpProvider`

adpi2:
If 2 entries contain the same class name, the first class should shadow the second class, because that the one loaded by the JDK `ClassLoader`.

Conicidentally that should fix https://github.com/scalameta/metals/issues/5807.


kasiaMarek (approved):
Doesn't seem related but some test are failing on the CI.

adpi2:
Yes the CI is always failing on this runner for a reason that I don't understand. Probably I should mute them on this runner specifically.

PR #834, opened 2025-01-23
improvement: Add location to failure to help mark it in VS Code

tgodzik:
I did a similar thing in https://github.com/scalameta/metals-vscode/pull/1571 but I wonder if it would safer here.

@adpi2 any idea if we can get a full path easily instead of just the name? Otherwise, it seems the regex is an ok solution.

### DIFF
--- a/modules/core/src/main/scala/ch/epfl/scala/debugadapter/Debuggee.scala
+++ b/modules/core/src/main/scala/ch/epfl/scala/debugadapter/Debuggee.scala
@@ -1,8 +1,10 @@
 package ch.epfl.scala.debugadapter
 
-import java.nio.file.Path
-import java.io.File
+import ch.epfl.scala.debugadapter.internal.SourceLookUpProvider
+
 import java.io.Closeable
+import java.io.File
+import java.nio.file.Path
 
 trait Debuggee {
   def name: String
@@ -19,4 +21,10 @@ trait Debuggee {
   def classPath: Seq[Path] = classPathEntries.map(_.absolutePath)
   def classEntries: Seq[ClassEntry] = classPathEntries ++ javaRuntime
   def classPathString: String = classPath.mkString(File.pathSeparator)
+
+  @volatile private[debugadapter] var sourceLookUpProvider: Option[SourceLookUpProvider] = None
+
+  private[debugadapter] def setSourceLookUpProvider(provider: SourceLookUpProvider): Unit = {
+    sourceLookUpProvider = Some(provider)
+  }
 }
--- a/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/ClassEntryLookUp.scala
+++ b/modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/ClassEntryLookUp.scala
@@ -37,6 +37,7 @@ private class ClassEntryLookUp(
     sourceUriToSourceFile: Map[SourceFileKey, SourceFile],
     sourceUriToClassFiles: Map[SourceFileKey, Seq[ClassFile]],
     classNameToSourceFile: Map[String, SourceFile],
+    sourceNameToSourceFile: Map[String, Seq[SourceFile]],
     missingSourceFileClassFiles: Seq[ClassFile],
     val orphanClassFiles: Seq[ClassFile],
     logger: Logger
@@ -48,6 +49,10 @@ private class ClassEntryLookUp(
     classNameToSourceFile.keys ++
       orphanClassFiles.map(_.fullyQualifiedName) ++
       missingSourceFileClassFiles.map(_.fullyQualifiedName)
+  }
+
+  def uriFromFilename(filename: String): Option[URI] = {
+    sourceNameToSourceFile.get(filename).flatMap(_.headOption).map(_.uri)
   }
 
   def classesByScalaName: Map[String, Iterable[String]] =
@@ -257,6 +262,7 @@ private object ClassEntryLookUp {
       sourceUriToSourceFile,
       sourceUriToClassFiles.toMap,
       classNameToSourceFile.toMap,
+      sourceNameToSourceFile,
       missingSourceFileClassFiles.toSeq,
       orphanClassFiles.toSeq,
       logger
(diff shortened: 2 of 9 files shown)


adpi2:
Thanks, that looks great!

> @adpi2 any idea if we can get a full path easily instead of just the name?

If you use the `SourceLookupProvider` you should be able to find the full source path from the class name.

Also I would appreciate if you could add a scripted test for it, similarly to [this is one](https://github.com/scalacenter/scala-debug-adapter/tree/4ee19303e42007933ead67ec4fb33ef835faa7cf/modules/sbt-plugin/src/sbt-test/debug-session/test-suite).

modules/core/src/main/scala/ch/epfl/scala/debugadapter/Debuggee.scala
```diff
@@ -19,4 +21,10 @@ trait Debuggee {
   def classPath: Seq[Path] = classPathEntries.map(_.absolutePath)
   def classEntries: Seq[ClassEntry] = classPathEntries ++ javaRuntime
   def classPathString: String = classPath.mkString(File.pathSeparator)
+
+  @volatile private[debugadapter] var sourceLookUpProvider: Option[SourceLookUpProvider] = None
```
tgodzik, line 25: This gets a bit hacky unfortunately, so not sure if we want to go that way. Alternatively, we could just add the name and resolve it in metals when forwarding the event.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/testing/TestSuiteEventHandler.scala
```diff
@@ -25,7 +26,8 @@ object TestSuiteEventHandler {
    * Provide a summary of test suite execution based on passed TestSuiteEvent.Results parameter.
    */
   def summarizeResults(
-      testSuiteResult: TestSuiteEvent.Results
+      testSuiteResult: TestSuiteEvent.Results,
+      sourceLookUpProviderOpt: Option[SourceLookUpProvider]
```
tgodzik, line 30: Bloop will not be able to provide this, so this becomes again more difficult. We could make it all public, but then it becomes a bit unmanagable without a a large refactor

tgodzik (commented):
I tried to go a bit further this direction, but it seems that getting source lookup to summarizing test event will require a large refactor than I hoped :sweat:

tgodzik:
Or maybe I could just switch to attaching stack trace? We already have utilities in Metals that can change that to an exact location. This would make this PR much simpler and we can intercept the message in Metals.

tgodzik:
So the alternative is in https://github.com/scalacenter/scala-debug-adapter/pull/837 and it will require only minor changes in Metals

modules/sbt-plugin/src/main/scala/ch/epfl/scala/debugadapter/sbtplugin/internal/SbtDebuggee.scala
```diff
@@ -68,7 +68,7 @@ private[debugadapter] final class TestSuitesDebuggee(
     s"${getClass.getSimpleName}(${target.uri}, [${tests.mkString(", ")}])"
 
   override def run(listener: DebuggeeListener): CancelableFuture[Unit] = {
-    val eventHandler = new SbtTestSuiteEventHandler(listener)
+    val eventHandler = new SbtTestSuiteEventHandler(listener, () => sourceLookUpProvider)
```
adpi2, line 72: This is really hacky.

It would be clearer to pass the `sourceLookupProvider` as a parameter of `run`.

```scala
override def run(listener: DebuggeeListener, sourceLookup: SourceLookupProvider): CancelableFuture[Unit] = {
  val eventHandler = new SbtTestSuiteEventHandler(listener, sourceLookup)
  
    ???
```

`run` is called by the debugger itself which always has an instance of `SourceLookupProvider`.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/Debuggee.scala
adpi2: I think it should not be a `var` but a normal parameter in the `run` method.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/testing/TestSuiteEventHandler.scala
adpi2: I am not sure to understand why Bloop could not provide an instance of `SourceLookUpProvider`. Bloop starts a debug server, which always creates an instance of `SourceLookUpProvider`.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/internal/ClassEntryLookUp.scala
```diff
@@ -50,6 +51,10 @@ private class ClassEntryLookUp(
       missingSourceFileClassFiles.map(_.fullyQualifiedName)
   }
 
+  def uriFromFilename(filename: String): Option[URI] = {
+    sourceNameToSourceFile.get(filename).flatMap(_.headOption).map(_.uri)
+  }
```
adpi2, line 56: I think we don't need this new method. You should be able to use `getSourceFileUri` instead which takes a class name, instead of a file name.

modules/sbt-plugin/src/main/scala/ch/epfl/scala/debugadapter/sbtplugin/internal/SbtDebuggee.scala
adpi2: Now I remember that `run` is called too early, before we construct the instance of `SourceLokkupProvider`. Maybe we should wait for the launch request before we start the debuggee process.

modules/core/src/main/scala/ch/epfl/scala/debugadapter/testing/TestSuiteEventHandler.scala
tgodzik: We would need to reimplement it in Bloop, or make SourceLookUpProvider not private.

modules/sbt-plugin/src/main/scala/ch/epfl/scala/debugadapter/sbtplugin/internal/SbtDebuggee.scala
tgodzik: What do you think about the other PR then? Might be simpler for us to calculate it in Metals

tgodzik:
Closing this one, since it's much more involved than I can work on right now. Alternative should be reasonably quick

PR #837, opened 2025-01-24
improvement: Add stack trace to failed results

tgodzik:
We can later check the first line using existing mechanisms in Metals to get the location


adpi2 (commented):
I hope it's not too verbose when there are a lot of failing tests. But otherwise it looks good to me.

tgodzik:
I also added the location, so that it's just a case of updating it in Metals

adpi2 (approved):

PR #862, opened 2025-04-15
fix: correctly load class with empty package from mirror

kasiaMarek:
fixes: https://github.com/scalameta/metals/issues/7401


tgodzik (approved):
LGTM! @adpi2 Do you still want to keep an eye on the repository or should we go ahead without your input?

PR #978, opened 2026-06-18
Add multi-output module support

zielinsky:
Introduce `MultiOutputModule` so clients can associate a single module with multiple class output paths.

### DIFF
--- a/modules/core/src/main/scala/ch/epfl/scala/debugadapter/ClassPathEntry.scala
+++ b/modules/core/src/main/scala/ch/epfl/scala/debugadapter/ClassPathEntry.scala
@@ -2,6 +2,7 @@ package ch.epfl.scala.debugadapter
 
 import java.nio.file.Path
 import java.net.URL
+import scala.util.Try
 
 sealed trait ClassPathEntry extends ClassEntry {
   def absolutePath: Path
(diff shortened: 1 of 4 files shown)


project/Dependencies.scala
```diff
@@ -38,7 +38,7 @@ object Dependencies {
   val scalaParallelCollection = "org.scala-lang.modules" %% "scala-parallel-collections" % "1.2.0"
   val scalaCollectionCompat = "org.scala-lang.modules" %% "scala-collection-compat" % "2.14.0"
   val sbtTestAgent = "org.scala-sbt" % "test-agent" % "1.12.12"
-  val scalaMeta = ("org.scalameta" %% "parsers" % "4.17.0").cross(CrossVersion.for3Use2_13)
+  val scalaMeta = ("org.scalameta" %% "parsers" % "4.15.2").cross(CrossVersion.for3Use2_13)
```
tgodzik: Why go back to an earlier version here?

tgodzik (changes requested):

project/Dependencies.scala
zielinsky: Ahhh, I didn't mean to push that change - sorry :/

tgodzik (approved):
LGTM!

PR #997, opened 2026-07-20
Release using Java 17

### LINKED ISSUE #993: Release workflow doesn't publish sbt 2.x plugin
When I added the sbt 2.x build of the sbt plugin in https://github.com/scalacenter/scala-debug-adapter/pull/986,
one thing could overlooked: The release-full workflow runs on [Java 8 in CI](https://github.com/scalacenter/scala-debug-adapter/blob/28f9db368f2509c709e0063871ec36f7eef1a639/.github/workflows/release-full.yml#L18), so it doesn't publish the sbt 2.x plugin as that requires Java 17+.

Example run: https://github.com/scalacenter/scala-debug-adapter/actions/runs/29107855733

Expectation: On releases, the sbt 2.x plugin should be published.

I'm not sure what's the best way to achieve this. Might be as easy as bumping CI to Java 17 (if the project itself wants to build using sbt 2.x, it anyway needs to do that).

yhefamly:
To publish the sbt 2.x plugin, a JVM 17+ is required in the release workflow. This bumps the JVM as needed and adds the release:8 flag on Scala 2.12 so that the bytecode is guaranteed to run on old JVMs.

Not fully tested as I can't run the release from local. Changes inspired by https://github.com/sbt/sbt-license-report/pull/226/changes solving a similar problem.

Closes #993.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -290,7 +290,7 @@ lazy val testOptionsSettings = Def.settings(
 
 lazy val scalacOptionsSettings = Def.settings(
   scalacOptions ++= onScalaVersion(
-    scala212 = Seq("-Xsource:3", "-Ywarn-unused-import", "-deprecation"),
+    scala212 = Seq("-Xsource:3", "-Ywarn-unused-import", "-deprecation", "-release:8"),
     scala213 = Seq("-Xsource:3", "-Wunused:imports", "-deprecation"),
     scala3 = Seq("-deprecation")
   ).value
--- a/build.sbt
+++ b/build.sbt
@@ -290,7 +290,7 @@ lazy val testOptionsSettings = Def.settings(
 
 lazy val scalacOptionsSettings = Def.settings(
   scalacOptions ++= onScalaVersion(
-    scala212 = Seq("-Xsource:3", "-Ywarn-unused-import", "-deprecation", "-release:8"),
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -290,7 +290,7 @@ lazy val testOptionsSettings = Def.settings(
 
 lazy val scalacOptionsSettings = Def.settings(
   scalacOptions ++= onScalaVersion(
-    scala212 = Seq("-Xsource:3", "-Ywarn-unused-import", "-deprecation"),
+    scala212 = Seq("-Xsource:3", "-Ywarn-unused-import", "-deprecation", "-release:8"),
```
tgodzik: Looks like this doesn't work for some version of 2.12. 


```suggestion
    scala212 = Seq("-Xsource:3", "-Ywarn-unused-import", "-deprecation", "-release", "8"),
```
could work

tgodzik (changes requested):

yhefamly:
macOS Java 17 fails one test with a timeout, I assume it's flaky as I can't see how the changes in this PR would cause this.

tgodzik (approved):
LGTM!

tgodzik:
Running the new release in https://github.com/scalacenter/scala-debug-adapter/actions/runs/29834716822 !

